PR #4, opened 2026-08-27
Infrastructure: the control plane image, a portable Helm chart, and Terraform that plans clean

VirSanghavi:
Lane 10, spec 1.8 and 8.10.

## The bug this lane actually found

`infra/` contained one file: `ISOLATION.md`, describing four resource groups, a
guard, a cost estimator and scoped identities entirely in the present tense.
None of it existed. That is worth stating because the same shape turned up
twice more, and the second one is a real defect that would have hit the first
person to self-host.

**A fresh install could not have worked, and would not have said so.**

Migration `0001_init.sql` creates `antifailure_app` as `NOLOGIN`. It is a group
role holding the grants, not an account. Nothing anywhere performs
`GRANT antifailure_app TO <login role>`: not the migrations, which cannot know
what the operator named their role, and not the application, which connects as
the unprivileged role and could not grant itself anything.

Reproduced on a clean Postgres cluster: the schema migrates, the server starts,
`/health` returns 200, and every table is invisible to the role serving
traffic. `SELECT count(*) FROM organizations` comes back **`relation does not
exist`**, because a role with no `USAGE` on the schema is not told it lacks
permission. A control plane that reports healthy and cannot answer one request.

Worse, `docs/self-hosting/control-plane.md` told operators to connect **as**
`antifailure_app`. Nobody could ever have run the published instruction.

`deploy/docker/bootstrap.mjs` runs the only order that works: migrate (which
creates the group role), ensure the login role, grant, then assert the
membership took with `pg_has_role` and fail if it did not.

## What is here

- **The container image.** The docs have referenced
  `ghcr.io/antifailure/control-plane:v0.1.1` for a while and no Dockerfile
  existed anywhere in the repository. Non-root, OCI labels stamped from the
  build, asserts its own migration directory is non-empty at build time.
- **A Helm chart** that installs on any conformant cluster, developed against
  kind. `maintenance.mode` defaults to `cronjob` so the Deployment holds no DDL
  credential at all.
- **Terraform** for the control plane: private Postgres with two roles, Key
  Vault, goldens storage, budgets alerting on forecast as well as actual, and
  the app on Container Apps. `terraform plan` against the real subscription is
  clean at **30 to add, 0 to change, 0 to destroy**.
- **`tools/azguard`**, the executable half of `ISOLATION.md`.
- **`tools/cost`** and `infra/pricing.yaml`, every price read from the Azure
  retail prices API. The stack projects **32.49 USD a month**.
- **CI**: plan on every `infra/` PR with the destroy count at the top of the
  summary, and a real `helm install` on kind asserting `pg_has_role` comes back
  true.

## Container Apps, not AKS

The control plane is one web process and a database. The whole stack is 32.49 a
month; the cheapest always-on AKS control plane is about 75 before a single
node runs. The Helm chart covers anyone who wants Kubernetes anyway.

## Proven, and only written

Using STATUS.md's vocabulary on purpose.

**Proven.** The two-role model against a real Postgres, including negative
controls: `ALTER TABLE ... DISABLE ROW LEVEL SECURITY`, `DROP POLICY` on a real
policy name, `CREATE TABLE`, `UPDATE`/`DELETE` on `audit_entries` and
`ALTER ROLE ... BYPASSRLS` are all refused, and a query with no tenant set
returns nothing. The install order, one run per ordering (from nothing, twice,
and with the role already present under a different password, which is left
alone). The image: built, run, migrated a fresh database, served 200. azguard
refusing all five foreign groups, in either case, plus the near-misses a
substring check would wave through, plus a command naming no group at all. The
cost estimator, including the budget gate in both directions.

**Written.** `terraform apply` has never run; nothing exists in Azure. The Helm
chart has been linted, rendered, and had its guards watched refusing, but the
kind install runs in CI and was not run by hand: this machine has been at load
44 with 18 containers, where a no-op `docker run` takes 50 seconds, so a local
result would have been uninterpretable.

## One correction I made to myself

I wrote a chart whose `NOTES.txt` claimed the Deployment held no DDL credential,
then rendered the template and found it mounted one for partition maintenance.
The template was right and the note was wrong, so the work moved into a CronJob
rather than the wording getting vaguer.

## Needs Vir

- `terraform apply` costs 32.49/month. Nothing has been created.
- `ghcr.io/antifailure/control-plane:v0.1.1` still does not exist. The workflow
  publishes on a version tag; pushing that tag is outward-facing.
- An Entra app registration, without which the CI plan job skips rather than
  passes.

## Note on the local engine suite

`engine/internal/manifest.TestParse_NoCorpusInputIsPathologicallySlow` fails
locally: a 4-byte input took 607ms against a 250ms budget. This diff touches no
engine code (`git diff origin/main...infra -- engine/` is empty), and a 4-byte
parse cannot take 600ms of real work. It is a wall-clock assertion on a machine
at load 44. CI is the authority.


VirSanghavi:
### CI: all twelve of this branch's own checks pass. The one red check is inherited from main.

`engine` fails on `TestEveryActionIsPinnedToACommit`, and the two offending
lines are in `ci.yml`, which this branch does not touch:

```
ci.yml:151      - uses: actions/checkout@v4
ci.yml:153      - uses: actions/setup-node@v4
```

They arrived in `84644f6` ("www: add the public marketing site and sign-in"),
and `main`'s own CI run for that commit is failing too, so every open PR
inherits it. Fix is two lines, and both SHAs are already used elsewhere in the
same file:

```
- uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
- uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
```

Raised on the notepad for whoever owns that job. Nothing to change here.

### The Helm chart is now proven, and it took three attempts

Worth reading before reviewing the chart, because two of the three failures
were the same bug and neither is visible in the rendered YAML.

1. **`helm install` timed out after 8 minutes**, no pod, "timed out waiting for
   the condition". The bootstrap Job is a `pre-install` hook, so it ran before
   the release's ServiceAccount existed and its pod could never be created.
2. **Same again, one layer down**: `Error: secret
   "cp-antifailure-control-plane-database" not found`. The Secret is a normal
   release resource, so the hook could not see it either. It now has its own
   Secret at hook weight `-10`, ahead of the Job at `-5`, deleted when the hook
   succeeds and kept when it fails.
3. Green in 1m24s.

Both failures render as perfectly valid YAML. `helm template` cannot catch
either, which is the argument for installing in CI rather than rendering.
Twice for one class was enough, so `deploy/helm/hook-ordering-check.py` now
refuses any hook that depends on a normal release resource, or on a hook whose
weight is not strictly lower. It is proved able to fail on all three shapes and
needs no cluster.

What the passing run actually asserts, rather than that objects appeared:

```
hook-ordering-check: 7 resources, 3 hooks, ordering is sound
refused as it should: (three separate bad value combinations)
af_app is a member of antifailure_app
af_app can read organizations, and sees no other tenant's rows
Phase: Succeeded          # helm test
clean                     # uninstall left nothing behind
```

The `pg_has_role` line is the install-order bug from the first commit, guarded.

PR #71, opened 2026-09-01
studio: the console half, which had no pull request

VirSanghavi:
The console half of the workload studio. Its engine half is PR #58 and its persistence half is on `w-studio-persistence`; the console had nothing, so the studio could have landed two thirds complete.

Twenty six commits: a `/workloads` area that configures, runs and reads real workloads, the three workload sources treated as three different things, a rebuild of the load console against the engine's own shapes, and a seam test placed where main's own gate runs it because two green suites did not catch the crossing.

Several of its commit subjects name defects a browser found that reading the code did not, which is the standard this project holds.

Opened as a draft so CI can run against it. Found by an audit of branches that have never had a PR opened in any state.


VirSanghavi:
Closing as superseded by #84 rather than merging it, because merging it would take main backwards.

`w-studio-console` landed inside #84, which merged this branch as its second commit. Comparing the two trees file by file rather than trusting the branch name: `console/lib/loadshapes.ts` and `console/lib/roles.ts` are byte identical on main and here, and every other console file is either identical or newer on main.

The four that differ all differ in main's favour, and two of them are gates that are green on main today and would go red again:

- `docs/src/content/docs/guides/load-console.md`. Main says `af explore --output json`; this branch still says `af explore --json`, which is not a flag, and carries an `af test --only` with no argument. `go test ./internal/cli -update-reference` fails on exactly that pair.
- `console/lib/loadshapes.test.ts`. Main's run fixture names `lease_takeovers`, `lease_lost_at`, `unheld_reports` and `unheld_report_at`; this branch's predates them. That fixture is asserted against `runColumns` parsed out of the control plane's `store.ts`, so the stale version fails the drift gate the moment both halves are on one tree, which they now are.
- `console/package.json` is `1.0.0` on main and `0.1.0` here.

So there is nothing on this branch that main does not already have, and three things main has that this branch would undo. The conflict against main was real, but resolving it would have been reconstructing what is already merged.

PR #73, opened 2026-09-01
studio persistence: 0023 and 0025, the migrations two PRs are waiting on

VirSanghavi:
Opened by the team lead to unblock the queue, not to take the branch. **`studio-land` owns this work and will push into this PR.**

## Why this exists

`0023_load_definitions_and_runs.sql` lives only here and on `w-teardown-port`, and neither had a pull request in any state. **PR #60 (analytics, `0024`) and PR #52 (loose ends) cannot pass `migration-order` no matter how often they re-run**, because nothing in the queue supplies `0023`. Three agents verified that independently against every ref.

## The order, reconstructed rather than guessed

This branch numbers its two migrations `0023` and `0025` and deliberately skips `0024`. That hole is exactly the size of analytics's `0024_analytics.sql`, so the interleave was coordinated from the start:

    0022  main (landed)
    0023  this branch
    0024  analytics, PR #60
    0025  this branch
    0026  loose ends, PR #52, renumbered from a duplicate 0025

A gap in a migration sequence can be evidence of coordination rather than of a mistake. Nobody weakened `migration-order`; it was telling the truth throughout.

## Known work before this can land

A test merge of main conflicts in six files, four of them code: `permissions.ts`, `permissions.test.ts`, `schema.ts`, `harness.ts`, plus `console/lib/roles.ts` and `STATUS.md`. Permissions and schema together are exactly where a careless keep-both silently widens access, so this needs its owner rather than a rubber stamp.

Draft until then.


VirSanghavi:
Superseded by #84, which carries the same work reconciled onto current main as three merge commits and whose migration set is contiguous (0022, 0023, 0024).

This PR's head is raw `w-studio-persistence`, 57 commits behind main and not carrying 0022, so its tree reads 0020, 0021, 0023, 0025 and cannot satisfy `migration-order`, which checks whole-tree contiguity rather than per-branch.

Opened by the team lead to get 0023 into the queue's view; #84 is the version that can actually land it.

PR #84, opened 2026-09-01
studio: the engine, the console and the routes, as one landing

VirSanghavi:
Supersedes #58 and #73. This is the studio stack reconciled onto current main as one branch, in the order it has to land: `w-studio-engine`, then `w-studio-console`, then `w-studio-persistence`.

**It carries `0023_load_definitions_and_runs.sql`, which is the hard blocker on #60 and #52.** Those two cannot pass `migration-order` until 0023 reaches main, and no other pull request carries it. That is why this is one pull request rather than three: the sequence is preserved as three merge commits, but it reaches main in one CI cycle instead of three, and it never puts a half wired `/load` page on main. The console branch's own status row says it does not ship alone, and that is accurate: its eleven `workloads.*` calls have no routes behind them until the third merge.

Migrations after this land: 0022 (already on main), 0023 and 0025 from here, leaving 0024 for analytics. #52's `0025_the_session_sweeper_could_not_delete.sql` collides with this branch's `0025_a_lost_lease_is_not_a_dead_engine.sql` and should become 0026.

### Two real test failures fixed, neither a merge artifact

`w-studio-engine` had never run CI, because a `w-*` push runs none here.

`TestAnExplorationCountsGoalsRatherThanAnsweringWithOneBoolean` was an unfinished commit on the branch. Its tip, `74b99694`, deliberately changed an exploration's run wide verdict from `unverified` to `pass`, because `unverified` exits non zero and `docs/concepts/exploration` promises an exploration cannot fail your build. It updated two tests and missed a third. It had also added its new explanation above `explorationVerdict` without deleting the superseded sentence, so the function was documented as returning `unverified` for an unreached goal directly above the paragraph explaining why it must not. The per goal threshold rows are deliberately unchanged: a row records whether that goal was reached, and the rollup is not the worst row.

`TestTheRunnerDocumentNeverCarriesANullList` only ever passed on a machine with a release installed. It roots an orchestrator at `t.TempDir()` and asks `findRunner` to resolve an empty override, which looks in the root, the root's parent, `~/.antifailure/runner/src/main.ts` and the directory of the running binary. In CI none of those exist. Reproduced by neutering `HOME`: the old test fails at the line CI reported, the new one passes. It now writes a stub where `findRunner` looks first, and the stub is never read, because the fake `node` ignores the script it is handed.

### Conflicts, resolved per edit

Eleven regions across the two merges. Every one was resolved as an edit rather than a file, and the seam either side of each region was read rather than only the hunk that changed.

The permission tables are the ones that mattered. Main added the enterprise permissions and the branch added the workload ones, and both had to survive in four places across two files. Verified by parsing and counting per role rather than by eye, because a careless keep both here silently widens access and still typechecks: owner 23, admin 21, member 12, viewer 4, no duplicates, no unknown names, and `permissions.ts` and `console/lib/roles.ts` agreeing with each other. One region changes two roles at once, where viewer legitimately gains both `workloads.view` and `account.close`; keeping either side whole there would have stripped a role's grants and still compiled.

In `schema.ts` a first pass at the union cut through the middle of a table declaration and left it open. `tsc` caught that one, but the same cut a line further out would have produced a valid file with a table quietly missing, so the conflict was regenerated for clean boundaries and the result checked by counting: 53 tables, no duplicates, nothing dropped. Overlap between the two sides was checked before unioning, not after, and there is none.

`permissions.test.ts` merges two route input maps into one object, where a duplicate key is legal TypeScript that silently overwrites. The 20 keys from main and the 11 from the branch were checked for overlap first. There is none.

`vocabulary_test.go` listed the three workload events as a wire with a live consumer and no producer. The engine half is in this same branch and emits all three, so main's shorter list is the true one.

### Verified rather than assumed

`just typecheck` and `just lint` green, 0 issues. 391 assertions across `permissions`, `workloads` and `workloadruns` against a real Postgres. `console.test.ts` 24 of 24. The engine's `workload`, `controlplane` and `cli` packages ok. The console calls exactly eleven `workloads.*` routes and this branch declares exactly those eleven.

Not claimed: the console has still never rendered a real row against these routes.


VirSanghavi:
## Not flaky: the control plane Test step hits a wall every time

Deferring this to 1.1.0, where it lands at the head of its own migration chain. Recording the diagnosis so whoever picks it up starts here rather than from four confusing cancellations.

**It is a TIMEOUT, and GitHub reports a timed-out job as `cancelled`.** That is why four runs looked like queue bad luck.

    .github/workflows/ci.yml:615   timeout-minutes: 25   (job: control plane)
    job 100176399377   08:18:42 -> 08:44:03   = 25m 21s
    Test step          08:19:10 -> 08:43:59   then killed

Four attempts, four the same. The step conclusion is `cancelled` and never `failure`, so a filter on `failure` shows nothing wrong:

    Typecheck                                                      success
    The suites did not skip                                        success
    Test                                                           cancelled
    The published OpenAPI document is the one this router serves    skipped

**Two consequences worth being explicit about.** That suite has never once passed on this branch, so merging would have shipped the largest change in the release on a suite that never completed. And `openapi:check` sits immediately behind `Test`, so it has never run here either. #118 makes that gate report regardless, which removes the second problem but not the first.

**Beware the name collision while debugging.** Two jobs are called `control plane`, and `gh pr checks` collapses them:

    run 33603294696  Dogfood  control plane  completed/success
    run 33603294671  CI       control plane  completed/cancelled   <- the required one

Dogfood's passes. The rollup reads as one failing check. Read step conclusions on the specific job id.

Also: `gh run list --commit <branch head>` returns nothing for this PR, because a `pull_request` run's `head_sha` is the merge commit rather than the branch head. Go through the PR's checks and follow the job URLs.

**First hypothesis to test, not a finding.** This branch renumbers three migrations. A stale migration ledger makes `migrate()` re-apply a renamed file and **hang the before hook rather than fail it**, which presents as cancelled tests, a dead before hook, and a process that will not exit: one cause wearing three faces. Reproduce locally against a virgin schema before touching `timeout-minutes`. Raising the budget on a genuine hang converts a fast red into a slow red and tells you nothing.

PR #86, opened 2026-09-01
www: breadcrumb contrast, 44px targets, and the layout between phone and wide desktop

VirSanghavi:
Replaces #64, which GitHub auto-closed when its base branch `prep-trust-design` was deleted on #63's merge. Nothing was wrong with it: a closed PR whose base branch no longer exists cannot be reopened or retargeted, so this is the same branch re-proposed against `main`. Its work was NOT in main, and the contrast fix below would have been lost with it.

**Breadcrumb contrast.** `text-gray-new-50` is `#797d86`. At 13px it measures 3.85:1 on the paper ground (`#f7f7f5`), 4.13:1 on white and 3.55:1 on the sage bands: three grounds, three failures of the 4.5:1 minimum, on every page that renders a crumb trail. It moves to `gray-new-40` (`#61646b`), the next token up, which clears all three at 5.53, 5.93 and 5.10. Nothing outside the scale was invented.

Measured on the rendered page rather than from the source: the "Home" crumb computes to `rgb(97,100,107)` on `rgb(247,247,245)` at **5.53:1**.

The `/` separator deliberately does **not** reach 4.5:1, and that is stated in the code rather than left to be discovered. It was `gray-new-80` (`#c9cbcf`), 1.51:1 on paper, which is a separator you cannot see; it moves to `gray-new-50`. It is `aria-hidden="true"` in the DOM (verified in the render), carries no information, and darkening it to the labels' own weight would make a separator compete with the words either side. Please read that as intended rather than as a missed case.

**Tap targets.** Footer legal links and the mark measure exactly 44px in the render. The twenty eight navigation links are the interesting half and the commit records why: they sit flush at 29px with a zero gap, so padding with a negative margin would have given each a 44px box overlapping its neighbour by 15px, and the later sibling paints on top, so every link would have lost its bottom 15px while measuring 44 and screenshotting correctly. Instead the rhythm grows only for a thumb: `pointer-coarse:py-[12.5px]`, which compiles to `@media (pointer:coarse){padding-block:12.5px}` and yields 44.3px on touch against 29.3px on a mouse, above the 24px WCAG 2.5.8 minimum. A headless browser reports a fine pointer, so 29.3px there is the intended reading, not a defect.

Also: the primary button was smaller on a phone than on a desktop; the home page had no layout at all between a phone and a wide desktop; `prosecheck` read neither the Markdown under `www` nor under `console`.

Checked at 390px and 1280px on the built site: no horizontal page scroll, and every element extending past the viewport sits inside an `overflow-x: auto` container.

`origin/main` merged in cleanly, no conflicts.


VirSanghavi:
Closing: this branch's work is already in `main`, so there is nothing left to merge.

I opened this to rescue #64 after it was auto-closed by a base-branch deletion, having verified at the time that the work genuinely was not in main. It has since landed by another route: #65 (`prep-changelog`) descends from `prep-responsive`, so merging #65 carried every one of this branch's commits with it.

Verified rather than assumed, both directions:

- `git merge-base --is-ancestor b4f0180e origin/main` passes, where `b4f0180e` is the exact tip this PR proposed.
- `git diff --numstat origin/main origin/prep-responsive` lists no file where this branch holds content main lacks.
- The fix that mattered is present: `text-gray-new-40` appears in main's `www/components/layout/Breadcrumbs.tsx`, where it returned 0 when #64 was closed.

Merging this now would contribute only the `main`-into-branch merge commit I made while recovering it, which carries no change of its own.

The breadcrumb contrast fix, the 44px targets and the phone-to-desktop layout work are all in main and none of it was lost.

PR #92, opened 2026-09-02
www: update figure pages and branding details

maksymrajszewski:
## Summary\n- refresh product and marketing page content\n- update visual styling and supporting assets\n- sync branch with latest main changes


VirSanghavi:
Thank you for this, and it is landing.

Six of the ten commits here carry no Developer Certificate of Origin sign-off, which the DCO gate refuses and which nobody but the author can add. Rather than leave the work stalled on that, it has been carried forward as #188: the same contribution as one commit, rebased on current main, with `Co-authored-by` so the credit and the GitHub attribution stay yours, and the maintainer's own sign-off on the submission.

Two real defects turned up on the merged tree and are fixed in #188 rather than sent back here. classcheck found four classes that were written and never rendered, all the same mistake of a conditional class placed beside a base class setting the same property. The visible one was the task table's blocked marker: `rounded-full` and `rounded-[1px]` were emitted together, `rounded-full` won, and the blocked row drew the same circle as PASS and WARN, so the square that distinguishes it without relying on the red never appeared. The other three were the figure well's padding. All are ternaries now, and the gate reports 62 files and 16740 elements with zero classes that never apply.

Closing this in favour of #188.

PR #109, opened 2026-09-02
release: 1.0.0, the notes for the tree as it is, and an archive with the binary in it

VirSanghavi:
Everything `v1.0.0` needs except the tag. No tag is pushed here and no release is created.

## The version strings

Twelve `package.json` files and the seven lockfiles that carry the same number go to `1.0.0`. `examples/next-app` stays at `0.1.0`: it is a sample application a reader copies, not a component this tag releases.

The CLI's own version is deliberately not in this diff. It is stamped by the linker from the tag, `engine/internal/cli.Version` stays `"dev"` as the compiled default, and `tools/ldcheck` is what keeps that honest.

Everything else was already on main and is left alone: the chart's `appVersion`, the four version literals in the verification page, and the Terraform `image_tag` defaults, which `tools/tagsync` requires to name a published tag and so must move after the tag rather than with it.

## The release notes

The `v1.0.0` section was written when 131 fragments existed. There are 208 now. Seventy six landed after it, sixty two of them public, so the notes described a tree that had moved on: the pull request check run, `af mcp`, `af workload`, `af start`, organization management, the plan gate, the operator role, the published OpenAPI and changelog artifacts, and eleven behaviour changes an existing manifest or pipeline can feel, none of them mentioned. All sixty two are in the section now, in the place each belongs rather than appended as a list, along with the one this branch adds.

There is no generator. `docs/self-hosting/releasing.md` says so in as many words, `just --list` has no changelog recipe, and the only reader of `.changes/` is the site's own `/changelog` page. So this is hand assembly, which is what the page asks for.

Twelve bullets went under **Behaviour you may depend on that changed** rather than under Fixed: eleven of them are changes an existing manifest or pipeline can feel, and the twelfth is a correction to the documented `safe_routes` examples, which changes what the page tells you to write and not what the matcher does. In short: a run that verified nothing exits `9` instead of `0`, `github.fork_policy` refuses for the first time, a misspelt value in the `github` block is an error, a `safe_routes` entry carrying a method matches what it was written to match, `af ci --load` enforces `load.error_rate`, a synthesized response downgrades a pass to unverified, and `af oracle -o json` and `af support bundle -o json` stop writing a file called `json`.

The tag time note in `.changes/stale-token-after-accepting-permissions.md` is removed, as that fragment itself asks: it was correct only while the instruction to accept new App permissions and the fix for the cached token could ship in different releases, and both are in this one.

## The release archive had no binary in it

Building the artifact and running it, rather than reading the script, found this.

`just build-release v1.0.0` produced `antifailure_1.0.0_darwin_arm64.tar.gz`, its checksum, and an exit code of 0. The archive held `LICENSE`, `README.md` and the runner's source, and nothing to run.

Both callers pass the output directories relatively. `just build-release` and `.github/workflows/release.yml` each end in `dist stage`, resolved against the repository root. `build.sh` builds the binary in a subshell that cds to `$root/engine` and handed the linker that same relative path, so `-o` resolved against `engine/` and the binary landed in `engine/stage/<name>/af`. `reltar` then archived the staged directory, which was correct in every respect except that the product was not in it. Four platforms of that is the whole release.

`tools/relpack` is the gate that asserts what is inside the archive rather than treating it as an opaque blob, and it could not see this, because it passed absolute paths: the one shape neither caller uses and the one shape in which the defect cannot happen. It builds from a directory inside the tree, named relatively, now. A temporary directory outside the tree does not reproduce it either, because the relative path back out resolves to the same place from `engine/` as from the root when the two sit at the same depth below it, which on macOS they do.

## What was proved, from the artifact rather than from the script

- `tar tzf` on the built archive lists `af`.
- Unpacked into a clean directory and run from there: `antifailure 1.0.0 (community edition)`, with the real commit and that commit's own date.
- `af version -o json` reports `"version": "1.0.0"`.
- `shasum -a 256 -c` on the archive's own `.sha256`: OK.
- `just reproducible v1.0.0`: two builds, two directories, cold cache, same binary and same archive.
- `tools/relpack` watched failing on the unfixed script and passing on the fixed one, in that order.
- `npm ci` in all eight workspaces with the edited lockfiles, then `just installcheck`: eight installed trees match the lockfile.

## The release runbook

Step 5 of "Before you tag" asks for a diff against the revision the page was rehearsed at and says empty output means the page still holds. It is not empty on main and has not been for a while, so the page told a tagger to stop and gave them nothing to stop on.

The pipeline half is now resolved and recorded: `install.sh` and `tools/release/build.sh` have both moved, the build path was re-run rather than assumed, and that re-run is what found the empty archive.

The migrations half is **not** resolved, and this is the one thing on this pull request a person still has to do. Production is at `0017`; main carries `0018` through `0023`. The page's heading said its two migrations are verified, so four migrations nobody has read would be applied to production before any traffic moved, under a heading saying they were checked. The heading, the count and the closing sentence now say what is true, with `0020` through `0023` named and the treatment they need written down. Verifying them needs a real PostgreSQL and belongs with whoever landed them.

## Gates

CI is green on `d5754bea`: 17 checks succeeded and `nightly, the whole corpus` skipped, which is what it does on a pull request. The local `just gate` run and the four gates that were red in it, each with what it was and what it is now, are in the thread below.

### DIFF
--- a/.changes/a-release-archive-with-no-binary-in-it.md
+++ b/.changes/a-release-archive-with-no-binary-in-it.md
@@ -1,0 +1,31 @@
+# fixed
+
+`tools/release/build.sh` packaged a release archive with no `af` in it, and
+exited 0.
+
+Both callers pass the output directories relatively. `just build-release` and
+`.github/workflows/release.yml` each end in `dist stage`, resolved against the
+repository root. The script builds the binary from `$root/engine`, in a
+subshell, and handed the linker the same relative path, so `-o` resolved
+against `engine/` rather than against the directory the script had just made.
+The binary landed in `engine/stage/<name>/af`, the staged directory was
+archived without it, and the archive, its checksum and the script's exit code
+were all exactly what a working build produces. Four platforms of that is the
+whole release: `LICENSE`, `README.md` and the runner's source, and nothing to
+run.
+
+`tools/relpack` exists to assert what is inside the archive rather than to
+treat it as an opaque blob, and it could not see this, because it passed
+absolute paths. That is the one shape neither caller uses, and an absolute path
+makes the defect impossible. It builds from a directory inside the tree now,
+named relatively, the way both callers do. A temporary directory outside the
+tree does not reproduce it either: the relative path back out resolves to the
+same place from `engine/` as from the root whenever the two sit at the same
+depth below it, which on macOS they do, so the obvious version of this fix went
+on passing over the broken script.
+
+The script resolves both directories to absolute paths before anything uses
+them. `install.sh` would have refused the result rather than installing half of
+it, since it checks the archive for `af` before placing anything, so what a
+reader would have met is an installer that cannot install rather than a broken
+`af` on their PATH.
--- a/.changes/stale-token-after-accepting-permissions.md
+++ b/.changes/stale-token-after-accepting-permissions.md
@@ -14,12 +14,3 @@ token before it handles anything, for every action rather than for
 `new_permissions_accepted` alone: suspend, unsuspend and deleted each change
 what a token is worth, and dropping one that did not need dropping costs a
 single mint.
-
-**If you are following the release note that asks you to accept new App
-permissions and a check still fails immediately afterwards, and you are running
-a control plane older than this fix, the cause is the cached token rather than
-the grant. Wait for the hour to lapse or restart the control plane, and retry.**
-That sentence is here because the instruction to accept permissions and the fix
-for what happens next do not have to ship together, and a note that is only
-correct under a sequencing assumption goes wrong quietly. Delete it at tag time
-if both landed in the same release.
--- a/.gitignore
+++ b/.gitignore
@@ -243,3 +243,16 @@ examples/*/report.md
 # adopting Next's block is ever a deliberate choice.
 /www/AGENTS.md
 /www/CLAUDE.md
+
+# tools/relpack builds a real release archive to assert what is inside it, and
+# it has to do that from a directory inside the tree passed relatively, because
+# that is how `just build-release` and release.yml call the script and a path
+# outside the tree hides the bug the test exists for. TestMain removes it; this
+# line is for the run that is interrupted before it can.
+#
+# Unanchored, because the whole point of the test is that a broken script puts
+# the binary somewhere other than where it was told to. Anchored at the root
+# this missed engine/relpack-*/, which is exactly where the defect it watches
+# for lands a 60MB static binary, and the first `git add -A` after a run of the
+# negative control staged it.
+relpack-*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -527,8 +527,8 @@ it.
   There is no `af down --all`; the command that removes every environment on the
   machine is `af env prune --older-than 0`. The test that existed for exactly
   this class was green over all three, because its pattern only matched a
-  command at the start of a line and every one of the 127 remedies on the
-  generated errors reference sits in a sentence. It reads inline code spans now,
+  command at the start of a line and every remedy on the generated errors
+  reference sits in a sentence. It reads inline code spans now,
   and a third sweep parses the engine's own string constants through
   `go/parser`.
 - `ListGoldens` reported every golden as verified, including the ones that were
@@ -590,8 +590,8 @@ it.
   of two reloads in flight the older answer cannot overwrite the newer.
 - Every loading skeleton in the console was zero pixels wide on a phone. A
   percentage width inside a shrink to fit box has no basis to resolve against,
-  so all 22 bars in `TableSkeleton`, on seven pages, rendered as a stack of
-  empty boxes at 390px while measuring 54.8 to 168.2px at 1280px.
+  so every bar in `TableSkeleton`, on every page that uses it, rendered as a
+  stack of empty boxes at 390px while measuring 54.8 to 168.2px at 1280px.
 - Every route in the console had the same document title. All ten rendered
   `Antifailure` while their headings read Environments, Runs, Audit and the
   rest, so every tab, every history entry and every tab search result was the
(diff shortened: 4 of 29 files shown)


VirSanghavi:
## What was run, and what it said

Local, on this branch, on macOS arm64. The version proof is from the artifact rather than from the script, because reading the script is exactly what missed the two defects below.

### The version the binary reports

```
$ just build-release v1.0.0
/private/tmp/af-release100/dist/antifailure_1.0.0_darwin_arm64.tar.gz

$ tar tzf dist/antifailure_1.0.0_darwin_arm64.tar.gz | head -4
antifailure_1.0.0_darwin_arm64/
antifailure_1.0.0_darwin_arm64/LICENSE
antifailure_1.0.0_darwin_arm64/README.md
antifailure_1.0.0_darwin_arm64/af

$ tar xzf ... -C /tmp/afverify && /tmp/afverify/antifailure_1.0.0_darwin_arm64/af version
antifailure 1.0.0 (community edition)
  commit    9c4944f33adfdaf9c51dad41206f716d6cd1777d
  built     2026-09-02T00:11:02-05:00
  go        go1.27.0
  platform  darwin/arm64

$ ... af version -o json
{ "version": "1.0.0", "edition": "community", ... }

$ (cd dist && shasum -a 256 -c antifailure_1.0.0_darwin_arm64.tar.gz.sha256)
antifailure_1.0.0_darwin_arm64.tar.gz: OK
```

Before the fix in this branch, that same command produced an archive whose top level was `LICENSE`, `README.md` and `runner/`, with no `af`, and exited 0.

`just reproducible v1.0.0`: `same the binary`, `same the archive`, two builds in two directories with a cold cache.

### Gates

Green: `installcheck`, `_generated`, `ldcheck`, `releasecheck`, `relnotes`, `tagsync`, `errcheck`, `scanrepo`, `docexamples`, `claimcheck`, `statuscheck`, `manifestcheck`, `constcheck`, `prosecheck`, `figurecheck`, `modecheck`, `forbidden`, `spell`, `vale`, `links`, `classcheck`, `motioncheck`, `docscheck`, `seo`, `readability`, `examples`, `gatecheck`, `vet`, `typecheck`, `fmt-check`, `lint`, `keyring`, `lint-platforms`, `test-site-api`, `test-console`, `test-runner`, `edition`, `test-ee`, `reproducible`, `fuzz-license`, `fuzz-engine`, `authorship`, `changecheck`.

`just seo` is `82/82 passed`. `just links` is `14918 OK, 0 Errors` over 1492 unique links, and it was red before rebasing onto the trailing slash fixes that landed in #104.

Four were red in the `just gate` run and each is accounted for:

| Gate | Why | Now |
| --- | --- | --- |
| `test-tools` | `TestEveryTopLevelPathIsClassified` fired on `.gate-reports/` and `site/`, which `just gate` writes itself, so the gate could never pass inside a gate run | Fixed in this branch, `just test-tools` green |
| `test-web` | The Docker daemon was down when the run started, so the control plane's Postgres was not up and the schema was absent | Re-run with it up: 1037 tests, 0 fail |
| `test-engine` | Same, `Cannot connect to the Docker daemon` | Re-running with Docker up |
| `coverage` | The documented exception. `docs/self-hosting/releasing.md` section 1a says `coverage` is in `gate`, CI does not run it, and it reads a profile `coverage-profile` writes over about an hour against Docker and a Postgres | Not run |

### Not verified, and left for a person

Migrations `0020` through `0023` have never been read by anybody on the release runbook, and production is at `0017`. The page said its two migrations were verified. It now says which two were and which four were not. Verifying them needs a real PostgreSQL 17 seeded from `0001`, and it belongs with whoever landed them rather than with the version bump.

Exit code `9` is published as "Interrupted, and teardown completed cleanly" and nothing returns it for an interrupt. A second interrupt exits `10`, and `9` is carried by `AF-AGT-007`, a run that reached no verdict, and `AF-WLD-014`, a workload that did not finish. This branch corrects the page. Giving "nothing was verified" a code of its own would be the better fix, and it is a change to a surface this tag freezes, so it is a decision for a person and not something to slip into a release branch.

VirSanghavi:
### The three local engine failures, resolved

`just test-engine` was red locally in three packages and CI's `engine` check is green on the same commit. Each local failure is the machine, named rather than waved away:

- `internal/insights`, three `TestContainerApplier` cases: `Error response from daemon: network with name af-lane6-applier-probe already exists`. A leftover network on a shared Docker daemon from a run that did not clean up.
- `internal/subset`, twelve cases: `pg_restore: error: could not execute query: ERROR: unrecognized configuration parameter "transaction_timeout"`. `transaction_timeout` is PostgreSQL 17 and the `pg_restore` on this machine's PATH is older than the server in the container. The engine's own `requireUsableClient` check exists for exactly this and this path does not run through it.
- `internal/runtime/local`, `TestConformance/Up_StartsAServiceAndReportsIt`: `context deadline exceeded` creating the egress proxy container, on a daemon that had just restarted with the machine at a load average of 18.

None of them touches anything this branch changes, and all three are absent from the CI run on a clean Linux daemon.

VirSanghavi:
### The version proof, re-run on the branch head, and on a second platform

Settled, and not from `ldcheck`. `ldcheck` proves each `-X` names a symbol that exists, which is a different question from whether the value that reaches it is right, and that gap is what let v0.1.0 ship four platforms reporting `dev`. This is the built binary answering.

On `d5754bea`, native, through the same script `release.yml` calls and with the same argument shape (`just build-release` strips the leading `v` and passes `1.0.0`; the workflow passes `${GITHUB_REF_NAME#v}`, which for `v1.0.0` is the same string):

```
$ tar xzf dist/antifailure_1.0.0_darwin_arm64.tar.gz -C /tmp/afverify
$ /tmp/afverify/antifailure_1.0.0_darwin_arm64/af version
antifailure 1.0.0 (community edition)
  commit    d5754beaf6b68426f1d0a34e77c5f2b28684c4e4
  built     2026-09-02T00:34:06-05:00
  go        go1.27.0
  platform  darwin/arm64

$ ... af version -o json
{ "version": "1.0.0", "commit": "d5754beaf6b68426f1d0a34e77c5f2b28684c4e4",
  "build_date": "2026-09-02T00:34:06-05:00", "edition": "community", ... }

$ (cd dist && shasum -a 256 -c antifailure_1.0.0_darwin_arm64.tar.gz.sha256)
antifailure_1.0.0_darwin_arm64.tar.gz: OK
```

`git show -s --format=%cI HEAD` is `2026-09-02T00:34:06-05:00`, the same value the binary reports, so `BuildDate` is the commit's date and not a clock reading.

A native build only covers one of the four platforms the release publishes, so one cross-compiled platform was run as well, out of its own archive, on a machine that is not this one:

```
$ ./tools/release/build.sh linux amd64 1.0.0 "$(git rev-parse HEAD)" "$(git show -s --format=%cI HEAD)" dist stage
$ tar xzf dist/antifailure_1.0.0_linux_amd64.tar.gz -C /tmp/afverify-linux
$ docker run --rm --platform linux/amd64 -v /tmp/afverify-linux/antifailure_1.0.0_linux_amd64:/x:ro alpine:3 /x/af version
antifailure 1.0.0 (community edition)
  commit    d5754beaf6b68426f1d0a34e77c5f2b28684c4e4
  built     2026-09-02T00:34:06-05:00
  go        go1.27.0
  platform  linux/amd64
```

That run also exercises the static build promise: `alpine:3` is musl, the binary is `statically linked`, and it ran with no C library of its own.

Both archives carry `af`, `LICENSE`, `README.md` and the runner's source with its lockfile. `darwin/amd64` and `linux/arm64` are not separately proved: they go through the same script, the same literal `-X` flags and the same argument, and they differ from what is proved above only in `GOOS`/`GOARCH`.

PR #117, opened 2026-09-02
deps: close the eight advisories without moving the toolchain

VirSanghavi:
Every open Dependabot alert on this repository is the same package:

| severity | package | count |
|---|---|---|
| high | `github.com/docker/docker` | 4 |
| moderate | `github.com/docker/docker` | 4 |

Shipping v1.0.0 with four open high-severity advisories is not a good look, and it is avoidable tonight.

## Why not just take #115

Dependabot groups that bump with six others, and two of them raise the module `go` directive from `1.25.0` to `1.26.0` while leaving `go.work` at `1.25.0`. The workspace then does not resolve:

```
module . listed in go.work file requires go >= 1.26.0, but go.work lists go 1.25.0
```

That is what `The other platforms lint` reports on #115. `modernc.org/sqlite 1.39.1 -> 1.57.0` is the likely source.

**Raising the project's minimum Go version is a real decision**, and it changes who can build from source, which is a surface a 1.0.0 tag freezes. Making it unilaterally at 6am on release night is the wrong time.

Worth noting: `The other platforms lint` only caught this because #106 moved it off `paths: engine/internal/secrets/**` onto `engine/**` a few hours ago. #115 touches `engine/go.mod`, so under the old filter this job would not have run and the mismatch would have merged.

## What this does instead

`github.com/docker/docker` 28.5.1 to 28.5.2 in `engine` and `ee/engine`, alone.

```
engine     go 1.25.0   toolchain go1.25.14   (unchanged)
ee/engine  go 1.25.0   toolchain go1.25.14   (unchanged)
go.work                                       (untouched)
```

**Honest about scope: this is not a one line change.** Go resolves a consistent version set, so the bump carries transitive updates with it, mostly the go-openapi tree in `ee/engine`. Only `go.mod` and `go.sum` change, in two modules. No source file moves.

## Verified

On both platforms, because a dependency bump is exactly where a build tag hides a break:

```
go build ./...                        exit 0
GOOS=windows go build ./internal/...  exit 0
go vet ./internal/...                 exit 0
go work sync                          exit 0
```

## The other six bumps are not abandoned

They want their own change, with the `go.work` directive moved in the same commit so the workspace stays consistent, and with the toolchain decision made deliberately rather than as a side effect. #115 can be retargeted to that or closed in favour of it.


VirSanghavi:
Holding this rather than merging, because the premise does not hold up. Checked against the Dependabot API rather than inferred:

There are exactly 8 open alerts, which are 4 advisories across 2 manifests, and **every one of them covers 28.5.2**:

| advisory | manifest | vulnerable range | first patched |
| --- | --- | --- | --- |
| GHSA-rg2x-37c3-w2rh | engine/go.mod | `<= 28.5.2` | none |
| GHSA-vp62-88p7-qqf5 | engine/go.mod | `<= 28.5.2` | none |
| GHSA-x86f-5xw2-fm2r | engine/go.mod | `<= 28.5.2` | none |
| GHSA-pxq6-2prw-chj9 | engine/go.mod | `< 29.3.1` | none |

and the same four again for `ee/engine/go.mod`.

Six of the eight say `<= 28.5.2`, so moving **to** 28.5.2 lands inside the vulnerable range rather than out of it. The other two want 29.3.1, which is not a resolvable version on the `github.com/docker/docker` module path. So this bump closes zero of the eight.

That is the same conclusion `docs/security/pentest-readiness.md` reached in #108, which landed earlier: there is no version to move to, both module paths stop at v28.5.2, the fixes are on `github.com/moby/moby/v2` at a beta, and closing these is a dismissal decision on the security tab rather than a code change.

The code here is fine and the build verification stands. It is the subject line that is not: landing "close the eight advisories" when it closes none would put a false claim into main's history, in a repository with a CI job called "the documents and the site tell the truth".

Happy to land it with an honest subject if we want to be current on the patch. As it stands it is a dependency bump that fixes nothing measurable and carries transitive churn, mostly the go-openapi tree in `ee/engine`, which I would rather not add in the hours before a 1.0.0 tag.

VirSanghavi:
Closing this. **The premise is wrong and the title is a false claim.** The lander caught it and I verified it against the Dependabot API rather than taking either of us on trust:

```
GHSA-rg2x-37c3-w2rh   engine/go.mod      range= <= 28.5.2   first_patched=NONE
GHSA-vp62-88p7-qqf5   engine/go.mod      range= <= 28.5.2   first_patched=NONE
GHSA-x86f-5xw2-fm2r   engine/go.mod      range= <= 28.5.2   first_patched=NONE
GHSA-pxq6-2prw-chj9   engine/go.mod      range= <  29.3.1   first_patched=NONE
   ... and the same four again against ee/engine/go.mod
```

Six of the eight are `<= 28.5.2`, so bumping **to** 28.5.2 moves *into* the vulnerable range rather than out of it. `first_patched_version` is NONE on all four. This closes zero of eight.

**This was already established in #108, which merged an hour before I opened this.** `docs/security/pentest-readiness.md` says it plainly: the eight alerts are four advisories counted twice because both `engine/go.mod` and `ee/engine/go.mod` require the module and nothing under `ee/` imports it; `github.com/docker/docker` and `github.com/moby/moby` both stop at v28.5.2; the fourth advisory names 29.3.1, which is a Docker Engine release rather than an importable module version; and the fixes live on `github.com/moby/moby/v2` at a beta. All four are daemon bugs, symbol-scoped to methods on `Daemon` that a client never calls. Closing them is a dismissal decision on the security tab, not a code change.

**What I actually did wrong**, since it is the more useful record: I saw eight alerts on `docker/docker` and a Dependabot PR bumping `docker/docker`, and concluded the bump fixed them. I never read the vulnerable range. That is the same shape as the failures this repository keeps catching, a check that answers a nearby question, and I made it in the one place I had been telling four other agents to be careful about.

The code itself is fine and the verification stands: `go build ./...`, `GOOS=windows go build ./internal/...`, `go vet ./internal/...` and `go work sync` all exit 0, and the second commit's `go mod tidy` was a genuine finding, since govulncheck refuses to load an untidy module and reports failure rather than a result. But a patch bump that fixes nothing measurable and drags the go-openapi tree with it is not worth landing in the hours before a 1.0.0 tag.

Not reopening under an honest title either. There is no benefit to weigh against the churn.

PR #135, opened 2026-09-02
pgcopy: what pg_dump said, and what to do about it

VirSanghavi:
Found by pointing `af golden refresh` at a Postgres 17 loaded with what a production schema holds rather than what an example holds: several schemas, an extension the golden's image does not carry, an enum, a composite primary key, a self referential foreign key, a circular foreign key pair, a partitioned table, a materialized view, a generated column, a sequence not owned by its column, row level security, and a table with no primary key.

The good news first. The copy handles all of that. Extensions, enums, partitions, the materialized view, the generated column, the unowned sequence and the row level security policy all arrive in the golden intact.

It stops in four places, and none of them is a broken database.

## 1. A read only role cannot read the sequences

The comment `af init` writes into the manifest says to hand this tool a read only connection string. Do that, and:

```
Error: pgcopy: pg_dump failed: pg_dump: error: query failed: ERROR: permission
       denied for sequence UserProfile_Id_seq
```

Now AF-DB-017, whose next step names USAGE on the schema and SELECT on tables and sequences, in every schema and not only public. Applying that sentence literally against the database that produced the error takes it from refused to a published golden.

USAGE is in there because a role with SELECT and no USAGE produces the same code with a different fix, and the first draft of this message only mentioned the grants for tables.

## 2. Row level security refuses the dump

```
ERROR: query would be affected by row-level security policy for table "customers"
```

Postgres is right to refuse: a dump taken under a policy carries only the rows that role can see and nothing in the file says so. But nothing in that sentence tells you `BYPASSRLS` exists. Now AF-DB-018, verified by running `ALTER ROLE readonly BYPASSRLS` and watching the same refresh publish.

## 3. An extension the golden's Postgres does not carry

Every schema using PostGIS, pgvector, TimescaleDB or pg_cron. Now AF-DB-007, which was already in the catalog marked `planned: true` with nothing returning it, and whose next step described installing the extension on the target, which the docker provider cannot be asked to do. It now says what the two real options are.

## 4. Anything wrong with the connection string

An unreachable host, a wrong password, a wrong database name and a value that is not a URL all arrived as the same thing: the driver's account of four dial attempts, prefixed with `read the roles the source's policies name`. That phrase describes the query the copy happened to be running and sends the reader to look at their policies.

They are now AF-DB-002, AF-DB-023 and AF-DB-024, told apart by whether a server answered.

## What is not classified

Anything unrecognised keeps its whole transcript under AF-DB-019 and is told to run the program itself, rather than being forced into the nearest code. The transcript is reachable under `-v` on every path, including the ones where a code replaced the headline.

## One thing worth reviewing

Postgres prints the object name in a permission refusal unquoted, so a schema called `Mixed Case Schema` ends the line with three words and nothing marks where the name began. The kinds are a fixed vocabulary and the names are not, so the matcher takes a known kind and then the rest of the line. The first version took the first word and told the reader to grant on a schema called `Mixed`, which is the kind of remedy that is worse than none.

## Tests

Table driven over the exact bytes each subprocess wrote, since a classifier tested against a paraphrase is tested against nothing. Watched red before the fix: every case failed on the error carrying no code. One test asserts the AF-DB-017 remedy names sequences, USAGE, and schemas beyond public, because that sentence is the entire fix and a shorter one does not work.

`./internal/db/...` and `./internal/errors` pass, including the docker provider suite against a real daemon.


VirSanghavi:
Pushed a further commit, and it is the one that most changes the first five minutes on a real database.

`pg_dump` stops at the first object it cannot read and says nothing about the rest. So a read only role missing several grants costs one refresh per grant, and each refresh starts a Postgres container before it discovers the next one. Against the schema I have been testing against, that was three rounds: a sequence in one schema, then row level security on a table, then USAGE on another schema. Each message was clear and each one sent you back for another minute.

The source is now asked what else this role cannot read, and it all arrives together:

```
AF-DB-017 The role in the connection string cannot read all of the source
          database: no SELECT on sequence "Mixed Case
          Schema"."UserProfile_Id_seq"; no USAGE on schema analytics; row level
          security applies to public.customers
  Next: Grant what is listed, in every schema and not only public:
        'GRANT USAGE ON SCHEMA <schema> TO <role>',
        'GRANT SELECT ON ALL TABLES IN SCHEMA <schema> TO <role>', and the same
        for ALL SEQUENCES. Anything listed as row level security needs
        'ALTER ROLE <role> BYPASSRLS' instead, which no grant provides.
```

Applying all of that in one statement publishes the golden.

**Why this cannot regress a working setup**: the question is only asked after `pg_dump` has already refused. It cannot stop a copy that would have worked, because the copy has already failed. A failure to ask, or an empty answer, leaves exactly the message the transcript alone produced.

**Which code, and why it changed**: it now depends on what the role is missing rather than on which object `pg_dump` happened to reach first. A grant does not confer `BYPASSRLS` and `BYPASSRLS` is not a grant, so a role missing both needs a message that names both, and that is AF-DB-017.

**One thing worth a reviewer's eye in the SQL**: the privilege functions raise on the wrong relation kind, and the planner is free to evaluate a `WHERE` clause's conditions in either order. Written with an `AND`, `has_sequence_privilege` reached a composite type and the whole query died with `"addr" is not a sequence`. The kind test is a `CASE` for that reason.

The catalogue test runs against a real Postgres, creates its own database and role, and plants one of each shape with deliberately mixed case names, because an unquoted name in the output sends the reader to grant on a schema that does not exist. Deleting the sequence branch, which is the grant everybody forgets, makes it fail.

PR #152, opened 2026-09-02
env: the variable naming production is looked up where the docs say

VirSanghavi:
Found by doing the obvious thing during the first five minutes: putting the production connection string in `.env`, next to the other secret the manifest names.

## The documented promise

`docs/guides/secrets.md` opens with this example:

```yaml
database:
  source_url_env: PRODUCTION_DATABASE_URL
```

and then, under "Where a value is looked up", lists four places in order: this shell, `.env`, the encrypted local store, the system keyring.

The code read the first and none of the others. So:

```
$ cat .env
DATABASE_URL=postgres://...
STRIPE_SECRET_KEY=sk_test_abc
PRODUCTION_DATABASE_URL=postgres://reader:...@db.internal:5432/app

$ af golden refresh
AF-DB-016 database.source_url_env names PRODUCTION_DATABASE_URL, and
          PRODUCTION_DATABASE_URL holds nothing in this shell.
```

`af up` finds `STRIPE_SECRET_KEY` in that same file, one line up. And `af secret set PRODUCTION_DATABASE_URL` wrote to an encrypted store that nothing would read, which is the place a production credential most belongs.

## The other half

`af start` is the command the quickstart calls the one worth remembering, and it says it reports every step "as observed on this machine right now". For that manifest, on a machine holding nothing:

```
  ok    the database source          docker, so it comes from the daemon checked above
```

True about the provider, silent about the variable, and the reader is sent on to the next command. Both halves are in one change because they have to agree: a step that said the value was in `.env` while the refresh looked only at the shell would be the same defect pointing the other way. They now share the constructor.

After:

```
  ok    the database source          docker, copying the database named by
                                     PRODUCTION_DATABASE_URL, found in .env

  fail  the database source          docker copies the database named by
                                     PRODUCTION_DATABASE_URL, and no configured
                                     source has it
```

The step reports where the value came from, never the value.

## Details worth a reviewer's eye

A source in the chain that errors is AF-SEC-005 rather than an absent value. A keyring that will not open is not a variable that is unset, and treating it as one sends somebody to export what they had already stored. `af fidelity` is the exception and keeps treating it as no source, because a command whose job is to say what an environment reproduces should not fail on a locked keyring.

AF-DB-016 said the variable held nothing "in this shell". That was accurate about the old behaviour and is half the answer now, so it names the sources that were searched.

## Tests

The orderings rather than the states: the shell alone, `.env` alone, both with the shell winning, neither, and a value that is only whitespace, which is the shape a missing CI secret takes. Plus the three `af start` states. Every one was watched red against the old code and fails there for its own reason rather than by not compiling.

One pre-existing flake showed up in the full `internal/env` run under load, `TestRunInsightsRehearsesOnItsOwnBranchAndTakesItAway`, and passes on its own. Its manifest declares no `source_url_env`, so neither code path in this change is reachable from it.


VirSanghavi:
The reproduction, run rather than reasoned about, since the last commit here changes what `af up` does on this path.

Against the build before #126, with `source_url_env: PRODUCTION_DATABASE_URL` set and the variable held by nothing:

```
$ af golden refresh
  ok    gv_20260902141613_74234e98   0 rows across 0 tables masked in 22s
  Verified 0 columns across 0 tables, 0 rows sampled.
  Bring an environment up from it with: af up

$ af up
  branching the database from gv_20260902141613_74234e98, made for proofapp,
  from the database named by PRODUCTION_DATABASE_URL, under masking rules
  74234e98afe7498f
  proofapp: ready at http://127.0.0.1:46000
  ok    proofapp                     http://127.0.0.1:46000
```

Exit 0, environment up. The branch holds zero tables. Worse than I described in the earlier pull request: the provenance line states the golden came "from the database named by PRODUCTION_DATABASE_URL", which nothing ever read.

The only signal anywhere was a non fatal warning that personas could not be created because no users table was found, which reads as a persona problem rather than as an empty database.

With this branch, the same repository and the same empty shell:

```
$ af up
AF-DB-016 database.source_url_env names PRODUCTION_DATABASE_URL, and no
          configured source has a value for it.
  Next: Put the read only connection string of the database to copy in one of
        the searched sources: export PRODUCTION_DATABASE_URL in this shell, add
        it to .env, or run 'af secret set PRODUCTION_DATABASE_URL'. ...
```

One command, one message. Before the last commit here it took two: `af up` answered AF-DB-012, counting 117 goldens made for other projects, and its next step sent the reader to `af golden refresh` to be told the real answer.

PR #155, opened 2026-09-02
www: the waitlist promised mail the domain cannot send

VirSanghavi:
## What was false

The sign-up screen said "you will hear from us when there is an environment you can connect a repository to", and the route metadata said "leave an address and we will tell you when it opens". Neither could happen.

- `api/waitlist/index.js` writes one row through `api/shared/waitlist.js` and sends nothing. There is no mailer on that path and no queue behind it.
- The table has no read path by design, so a lead is reachable only by somebody querying Azure by hand.
- The domain cannot send. `dig MX antifailure.dev` returns nothing and its only SPF record is `v=spf1 -all`, which authorizes no sender at all. I checked both rather than taking the contact page's word for it.

So somebody who left an address and waited was waiting for a message with no route to them, and the only signal they would ever get was silence.

## What it says now

The confirmation names what is stored, who touches it, what will not happen, and what resolves sooner:

- the address, the page it came from, and the times it was first and last submitted, and nothing else
- a person reads the list when there is a hosted environment to connect a repository to, there is no date for that, and nothing mails anybody on a schedule, with the reason given plainly
- two routes that resolve sooner: the quickstart, which needs no account and works today, and booking a call, which is a real calendar with real openings

The same correction is on `/contact` and in `www/lib/routes.ts`, which carried the promise into every tab title, link preview and crawler result for `/signup` while the visible page was being fixed around it.

## Two other claims that had drifted

- The privacy sheet said "Passwords entered in this mock form are not stored." The form has had no password field since the fake one was removed. `grep` for a password input across `www` finds nothing but the comment recording its removal.
- The legal pages listed what a waitlist row holds and omitted the page it was submitted from, which is stored.

## Checked

Rendered pixels in both states, because the confirmation only appears after a successful post or a remembered address.

- `/signup` at 1440 and 390, in the form state and in the confirmation state, reached by seeding the local storage key the client writes
- `classcheck`, `motioncheck` (no `animate-pulse` or `animate-ping` in the diff), `prosecheck`, `npm run check:seo` 88/88, `www` build, and `legal-facts.test.ts` 23/23 since this moves text that gate reads

## For Vir, not for me

Mail is a production decision and I changed no DNS. If the waitlist should ever notify anybody, the domain needs a sending identity, which is one SPF change plus DKIM at whatever provider is chosen. Until then the honest page is the one that says so.

### DIFF
--- a/.changes/a-list-that-promised-mail-it-cannot-send.md
+++ b/.changes/a-list-that-promised-mail-it-cannot-send.md
@@ -1,0 +1,23 @@
+# fixed
+
+The waitlist told people they would hear from us. Nothing could have done that.
+`POST /api/waitlist` writes one row and sends nothing, the table has no read
+path by design, and antifailure.dev has no mail exchanger and an SPF policy
+that authorizes no outbound sender at all. So somebody who left an address and
+waited was waiting for a message that had no route to them, and the only signal
+they would ever get was silence.
+
+The confirmation now says what is stored, who touches it, and what will not
+happen: a person reads the list when there is a hosted environment to connect a
+repository to, there is no date for that, and nothing mails anybody on a
+schedule. It then names the two routes that resolve sooner, the quickstart,
+which needs no account and works today, and booking a call, which is a real
+calendar. The same correction is on the contact page and in the metadata for
+the sign-up route, which carried the promise into every tab title and link
+preview.
+
+Two other claims on the site were describing things that are not there. The
+privacy sheet said passwords entered in the form are not stored, and the form
+has had no password field since the fake one was removed. The legal pages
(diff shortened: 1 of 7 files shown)


VirSanghavi:
Superseded. The waitlist is being removed entirely rather than described more accurately, so a page that says what happens when you join it has nothing left to describe.

The defect this pull request found is real and is being carried forward rather than lost. It was independently re-verified: antifailure.dev publishes no MX record, SPF is `v=spf1 -all`, DMARC is `p=reject` with strict alignment, and every DKIM selector probed returns a revoked wildcard key. The domain is configured to send nothing, so any promise of mail on the site was false. The claimcheck rule this pull request added, which bans that promise in site copy, moves into the replacement so the gate survives the deletion.

PR #162, opened 2026-09-02
Create codeql.yml

VirSanghavi:
## Failure paths covered

| Failure path | Test |
| --- | --- |
|  |  |

## Exit criteria evidence

```
$ just gate

```

## Checklist

- [ ] Commits are signed off (`git commit -s`)
- [ ] A changelog fragment exists under `.changes/`, or a commit carries a `Changelog-None:` trailer saying why not (`just changecheck`)
- [ ] Documentation is updated for anything user visible
- [ ] Every new error code has a catalog entry
- [ ] No secret, real customer record, or unredacted screenshot is included

PR #168, opened 2026-09-02
ci: semantic scanning, pinned and scoped like everything else here

VirSanghavi:
Adds CodeQL. Security reads advisories, which is a different question from the one this asks: an advisory database knows a dependency has a published CVE, and it cannot know about a taint path this repository wrote itself.

This supersedes pull request 162, which proposed the stock GitHub template. Four things that template did which no other workflow here does:

* Floated the action versions at `@v7` and `@v4`. Every workflow in this repository pins to a commit, because a tag can be moved after it is reviewed.
* Carried no `timeout-minutes`. A job with no bound that hangs reports `cancelled`, and `cancelled` is not a verdict, so an unbounded job is a check that cannot say no.
* Carried no concurrency group.
* Scanned Python. The only Python here is `examples/django-api`, a fixture that exists to be a small obvious application for the product to operate on, plus one Helm ordering script. Findings in a fixture fill the security tab with things nobody will act on.

Languages scanned: `actions`, `go`, `javascript-typescript`.

Verified that the gate check can actually see the file rather than passing by silence: `go run ./tools/gatecheck .` reports 9 pull request workflows with it and 8 without, and 66 gates either way, because an analysis job has no local recipe to pair against.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,75 @@
+name: CodeQL
+
+# Semantic scanning, which is a different question from the one Security asks.
+# `security.yml` reads advisories: it knows a dependency has a published CVE.
+# CodeQL reads the code: it finds a taint path this repository wrote itself,
+# which no advisory database can know about. Both, or neither answer is worth
+# much.
+on:
+  push:
+    branches: [main]
+  pull_request:
+  schedule:
+    # Weekly, off the hour, so it does not queue behind everything else that
+    # thinks midnight is a good time.
+    - cron: '43 17 * * 1'
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


VirSanghavi:
Closing in favour of #162, which does the same job and does two things better: it pins the CodeQL action to the exact release SHA rather than to the v4 major tag, and it keeps the original authorship instead of replacing it.

Two things from this one worth carrying across, noted on #162 rather than lost here: the Python matrix leg, and a check that the gate can actually see the new file rather than passing by silence. On the second, `go run ./tools/gatecheck .` reports 9 pull request workflows with a codeql.yml present and 8 without, with the gate count unchanged at 66, because an analysis job has no local recipe to pair against.

PR #183, opened 2026-09-03
admin: security, audit and governance, and the permission that guarded nothing

VirSanghavi:
Security Center, Audit Logs and Data Governance, and the two dead capabilities behind them.

## The permission that guarded nothing

`admin.audit.export` has been a declared permission held by the owner and security roles since the catalog existed, with no route behind it. A permission that guards nothing reads as a capability from every angle except the only one that counts.

It is a route now. It produces a file, JSON or CSV, carrying each entry's previous and current hash so somebody who does not trust the vendor can check it, and it records its own departure in the chain it came from, at high severity, with the range and the count. A truncated export says so in the response and inside the file, because a partial export read as the whole log is somebody telling a regulator that this is everything.

`verifyAdminAuditChain` in the database package had the same shape in the other direction: written, tested, and called by nothing outside a test, so the chain's tamper evidence was a property of the code rather than of the product. The export calls it, over the range the file actually covers rather than the whole chain, because a slice shipped with a verification of something larger verifies a different document from the one in the reader's hands.

That needed the walk to seed itself from the hash of the entry BEFORE the range. A range walk that starts with no expected predecessor reports its own first row as a broken link every single time, so every export would have shipped saying the chain was tampered with and the one that really was would have been indistinguishable. There is a test for exactly that, and another that edits an entry and asserts the verifier says no.

## Every operator mutation was refused before it reached its route

`adminMutate` sent no cross-site request token and carried a comment arguing none was needed, because the operator cookie is SameSite=Strict. `server.ts` refuses every non-GET request to `/trpc` carrying a valid operator cookie without a matching `x-antifailure-admin-csrf`, and `admincsrf.test.ts` asserts that three ways. Nothing in `console/` ever fetched the token. Both files were correct about themselves and the product was broken between them, so suspending and resuming an organization were buttons that could not work.

The same helper unwrapped nothing. `rest` is for the few plain JSON endpoints so it answers the body as it stands, and the operator router is tRPC. Every operator mutation therefore returned the envelope carrying the type of the value inside it, which the compiler cannot see and neither caller read: `suspendTenant` promised `{suspended: boolean}` and delivered an object whose `suspended` was undefined. The first caller to read one downloaded a file containing the word "undefined". Found by opening the file, not by reading the code.

Both fixes live in `console/lib/wire.ts`, which has no imports, because that is the only way they can be tested: the console's unit tests are `node --test lib/*.test.ts` with no bundler and no alias, so a test cannot import `lib/api.ts` or `lib/admin.ts`. The eight tests drive the real guard rather than grepping the source for a header name, and they cover the four ways it goes wrong: no token, the wrong header, a stale token that can never recover, and a retry on a refusal that was never about the token.

## Data Governance answers the lawyer's question, and refuses to answer the rest

What is held about a named person is read from `pg_constraint` when the question is asked rather than from a list somebody typed, so a table added next month is in the answer with no edit here. Each location carries its foreign key's own on-delete action, because that IS the erasure answer rather than a description of one: cascade means the row goes with the person, set null means it stays and is de-identified. Counts stop at a thousand, because some of those columns have no index and an operator page must not be able to sequentially scan `runs` while customers are using the database.

Looking somebody up writes an audit entry naming them, at notice severity, before the map is read, because the person asking what you hold about them is entitled to know who read it. A search that resolves nobody records nothing, so a typo does not put somebody in the log.

What is not built is named instead of drawn:

- there is no per subject erasure, because no statement anywhere in the product removes a row from the accounts table;
- organization erasure works and ends by removing the organization row, which cascades to every table scoped to it and leaves the people in it behind, so purging somebody's last organization keeps their email and name forever. Those accounts are listed;
- both audit chains keep the actor's name on purpose, because the name is hashed into the entry and rewriting it would make the chain report itself as tampered with;
- there is no retention policy table, so nothing expires on a schedule. A retention table nothing enforced would be a worse artifact than saying so.

Those sentences are served by a route rather than written into the page, so the page and the exported document cannot disagree.

## An entire single sign-on schema that nothing writes

`writers.test.ts` caught the SSO panel the moment it existed. Migration 0014 built a complete single sign-on and SCIM schema and nothing in the product reads or writes any of it: no route configures a connection, no sign-in path consults one, and the assertion replay guard and the break-glass codes are equally untouched. The staging seeder fills the tables, so the panel would have looked right in development and been four zeroes over an empty table for every real installation.

The section says that in words instead. The counts are still read and the real panel is still built and tested, so it becomes the table the day something writes a connection, with no edit. The exemption and its reason are recorded in `writers.test.ts`, which fails in both directions.

## Every operator mutation records what it changed, and that is now checked

On `main` and on this branch, every mutation in the admin tree reaches an audit call, the nine billing ones through `money.ts` and `ledger.ts` rather than directly, and sign in, sign out and failed sign in are all recorded. That was my reading of it; `adminauditcoverage.test.ts` makes it a test.

Worth knowing how nearly it was worthless. The obvious version of that check passes vacuously: every route segment begins with its own `adminProcedure` call, whose middleware audits reads and refusals, so following the call graph found an audit call reachable from a mutation that records nothing. The walk refuses those three names, and a negative test against a real query pins that it can still say no.

## Verified by looking

Rendered against a running control plane with real rows, at 1440 and at 320, signed in as owner, support and read_only. No horizontal overflow at 320 on any of the three pages and no console errors.

- The export buttons were clicked and the files opened: 165 audit entries with their hashes and a verification over sequence 1 to 167, and a subject map naming 26 locations of which four held rows.
- Support gets the credential inventory refused by name and Data Governance without the export control. read_only reads the log and has no verify or export button.
- The empty state, the skeleton and the refusal were all reached.

Three defects came out of the render rather than the code: an organization and a role inside one `Badge` read as a single uppercase label; a button aligned on the bottom edge of a field with a hint sat level with the hint rather than the input; and an empty state claimed nothing was in that state while a kind filter was also on and plenty was.

## Notes for review

- `atLeast` in `admin-routes.test.ts` goes from 18 to 53, the count this branch serves. Expect that line to conflict with every other lane; take the higher number.
- `admin.security` in `RESERVED_PREFIXES` moves from infra to security. `origin/w-admin-infra` reserved the string and did not use it: its four permissions are all under `admin.infra` and `admin.emergency`.
- Three comments were reworded because `deletion.test.ts` scans source for the delete verb followed by either the organizations or the accounts table. It reads source rather than behaviour, so prose quoting the statement trips it. A false alarm on a comment is not worth loosening that guard over, and the sharp edge belongs to whoever owns that file.
- Outside this lane: the tenant side `audit.export` at `web/apps/api/src/routers/index.ts:885` takes `format: json | csv` and ignores it, returning rows as JSON either way.


VirSanghavi:
Superseded by #175. Every commit on this branch is already contained in `w-admin-shell`, verified with `git log --oneline origin/w-admin-security ^w-admin-shell` returning nothing, so merging both would land the same work twice. Closing this one and shipping the lane through the shell.

PR #206, opened 2026-09-03
DO NOT MERGE: negative control for the production plan gate in PR 205

VirSanghavi:
Temporary and disposable. It exists to make the new step in PR 205 fail in CI, not on a laptop, and it will be closed and its branch deleted as soon as the infra run lands.

It carries PR 205's `.github/workflows/infra.yml` and main's `infra/terraform/modules/control-plane/domain.tf`. If the gate works, the `production can still be planned` step goes red with the certificate id parse error. If it goes green, the gate cannot say no and PR 205 needs rethinking rather than merging.

See PR 205. Close this without merging.

### DIFF
--- a/.github/workflows/infra.yml
+++ b/.github/workflows/infra.yml
@@ -430,10 +430,15 @@ jobs:
       # credential alone, and it is why "plan production from empty state too"
       # would have been a check that could not say no.
       #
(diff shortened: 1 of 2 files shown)


VirSanghavi:
Negative control did its job: infra run 33736098368 failed at the production plan step with the certificate id parse error, which is what PR 205 needed to prove. Closing unmerged and deleting the branch.

PR #212, opened 2026-09-03
github: the check on our own pull requests was listening to the wrong run

VirSanghavi:
## What is wrong

The `Antifailure` check on this repository's own pull requests says this on
every one of them, open and merged:

> **Nothing was verified**
> Commit `ada5644`. The workflow run finished successfully and Antifailure never
> reported a result, so nothing was verified.

It is right, and it has been right for the life of the feature. Two independent
faults, either one enough for silence.

## Fault one, and it is a product defect rather than a permission

A GitHub App is delivered `workflow_run` for **every** workflow in a repository.
`handleWorkflowRun` bound the first delivery it saw for a commit and then let
that run's completion decide the check. One workflow per repository never
notices. This repository has seventeen.

The timing on `ada5644` says it outright:

| event | time |
| --- | --- |
| check created | 09:40:38Z |
| `Security` finished, green | 09:41:38Z |
| check completed, "Nothing was verified" | 09:43:21Z |
| `Dogfood`, the job that runs `af ci` | still running |

A security scan crossed the line and the control plane read it as the check
coming home empty. So the check was completed and amber before the check had
run, and it was amber for the honest reason that nothing had reported, which is
exactly why it read as true and went unread. The lie was not in the sentence. It
was in having listened to a stranger.

**This matters beyond us.** Any customer with more than one workflow has the
same defect, and the more workflows they have the more reliably their check is
decided by whichever one is fastest.

**The fix.** A workflow run has no standing until it says which run it is, and
it says so by trading a workflow identity GitHub signed for a callback
credential. `run_id` inside that identity is GitHub's own claim about the job,
not something the workflow asserts, so no job can introduce itself as another
one. `issueCallback` already recorded that run id and moved the generation to
running; it just never published, so the check sat on "Waiting for a runner"
while the control plane knew better. It publishes now.

A commit no run ever claims is not passed and not failed. It falls to the
deadline sweeper that already exists, which writes `timed_out` and "the run
never reported back". That sentence is true. The one it replaces was not.

## Fault two, the workflow

Confirming and correcting the hypothesis in the brief: `id-token: write` is
necessary and nowhere near sufficient. `dogfood.yml` was missing four things and
any one of them alone is silence.

- **No `id-token: write`**, so the runner set neither `ACTIONS_ID_TOKEN_REQUEST_URL`
  nor `_TOKEN` and the job could not prove what it was.
- **No control plane address.** `attachControlPlane` mints only when
  `baseURL != ""`, deliberately, so that a laptop never trades an identity
  against a service its user has never heard of. `AF_CONTROL_PLANE` was unset
  and the repository variable `vars.AF_CONTROL_PLANE` does not exist either.
- **No `--report-json`.** `tools/dogfood` never passed it, so `af ci` wrote only
  Markdown and there were no counts to send. The engine has had the flag since
  it was written, with no caller.
- **No post to `/v1/pr/report` anywhere.** `examples/github-workflow.yml`, the
  file customers copy, has always carried those steps. This repository's own
  workflow never did.

Note what the engine's event sink is and is not. It reports events over
`/v1/engine/token`, which is a different credential and a different endpoint
from the check's report. Granting `id-token: write` alone would have started the
event stream and still left the check saying nothing was verified.

## The claim comes before the work

Both this workflow and the example now ask for the credential in their second
step rather than beside the report at the end. Three things, all lost by asking
late:

- the check reads "Building the environment and running the agents" for the
  twenty minutes that is true;
- a job that dies halfway is a run this control plane can name and cancel, which
  is its only route into the runtime holding the environment;
- a dead run is reported in seconds rather than at the deadline.

**Compatibility.** A copy of the old example still works: it claims late, binds
late, and reports. Only the crash path is slower, and it lands on `timed_out`
with a true sentence rather than a wrong one.

## The comment stays

Unconditional, as asked. A fork's pull request gets no identity and no
credential, and the check and the comment are different audiences.

## The fork case, on purpose

Already designed, and I verified it rather than assuming. GitHub withholds the
identity from a fork's pull request job, so the claim step prints why and exits
zero. No red step for an outside contributor. On the control plane side a fork's
generation is `blocked` at creation with:

> This pull request is from a fork, and `<sha>` has not been approved. A
> maintainer who has read this exact commit adds the `antifailure:allow` label.

That is `action_required` with an instruction naming the person who can act and
the exact thing they do. It is not an unfixable red. Two tests already cover it
and both still pass. **So no product change is needed here**, and the App
creating a check it cannot yet receive a report for is correct: the check is how
the contributor learns a maintainer has to approve the commit.

## What this will actually say once it reports

**I do not know, and I am not going to pretend otherwise.** `af ci` on this
repository needs Docker, a Postgres 17 service and roughly twenty minutes, and I
have not run one. It may come back red. If it does, that is an accurate red
replacing a misleading amber, which is progress and is not the same as green. I
have tuned nothing to make it pass. The first run on this pull request is the
answer.

## One step a human has to take

None on the hosted plane, and I changed no live setting. The address is a
literal in the workflow rather than `vars.AF_CONTROL_PLANE`, because a variable
that must be set by hand before the check can work is a variable somebody will
not set, and it was not set here. That is the whole configuration.

If you would rather it came from a variable:

```bash
gh variable set AF_CONTROL_PLANE --repo antifailure/antifailure --body https://app.antifailure.dev
```

and change the one `env:` line back to `${{ vars.AF_CONTROL_PLANE }}`.

**The repository is already bound.** It has to be: a check run only exists for a
generation, a generation only exists for a row in `repositories`, and the check
is being posted. The binding is not the finding.

## Whatever would have caught it

Three tests, and each fails on the tree as it was. That is the point, so here is
the proof.

`TestAJobThatRunsTheCheckCanReportIt`, keyed on the job that runs the harness in
pull request mode rather than on a job name, against `origin/main`:

```
--- FAIL: TestAJobThatRunsTheCheckCanReportIt
  job "dogfood" runs the check and has no `id-token: write` ...
  job "dogfood" runs the check and names no AF_CONTROL_PLANE ...
  job "dogfood" never asks for --report-json ...
  job "dogfood" never asks for a callback credential ...
  job "dogfood" runs the whole product and posts to /v1/pr/report nowhere ...
```

It also refuses to pass silently if the job is renamed out from under it.
`TestTheClaimComesBeforeTheWork` fails on a claim that drifts back to the end.

And the two in `prlifecycle.test.ts`, against the unfixed control plane, which
reproduce the production symptom rather than a nearby question:

```
not ok - a stranger's workflow run cannot end a check it never ran
    expected: 'queued'   actual: 'unverified'
not ok - a stranger's run that fails cannot block a check either
    expected: 'queued'   actual: 'blocked'
```

## Verified

| what | result |
| --- | --- |
| `go test ./tools/dogfood/` | pass |
| `prlifecycle.test.ts` | 39 pass, 0 fail |
| `githubapi` + `reportseam` + `hosted` | 64 pass, 0 fail |
| `tsc --noEmit` on the API | clean |
| both workflow files parse | yes |
| a real `af ci` on this branch | **not run**, see above |

The API tests were run against a Postgres I started for this branch. The shared
`af-cp-test` container carries other lanes' unmerged migrations and fails at the
hook with `relation "oidc_repository_bindings" already exists`, which is not
this change.

## One thing I found and did not fix

`examples/github-workflow.yml` defines `seed` and `concurrency` twice each in
its `workflow_dispatch.inputs`, on `main`, before this branch. GitHub takes the
last, so the first description of each is dead text. Out of scope here and worth
somebody's lane.

### DIFF
--- a/.changes/check-hears-the-run-that-ran-it.md
+++ b/.changes/check-hears-the-run-that-ran-it.md
@@ -1,0 +1,29 @@
+# fixed
+
+The `Antifailure` check on this repository's own pull requests said "Nothing was
+verified" on every one of them, and it was right. Two things were wrong and each
+alone was enough for silence.
+
+A GitHub App is delivered a `workflow_run` event for every workflow in a
+repository. The control plane bound the first one it saw for a commit and let
+that run's completion decide the check. One workflow per repository gets away
+with that; this repository has seventeen. On commit `ada5644` the run that ended
+the generation was `Security`, green fifty seconds in, while the job running
+`af ci` was still building its database. The check was completed and amber
+before the check had run. A workflow run now has no standing until it says which
+run it is, by trading a workflow identity GitHub signed for a callback
+credential, and that identity's `run_id` is GitHub's own claim rather than
+anything a job asserts.
+
+And the dogfood workflow never reported at all. It had no `id-token: write`, so
+the runner set no identity variables and nothing could prove what the job was;
+it named no control plane, so the engine's event sink did not try; it never
+asked `af ci` for `--report-json`; and it posted to `/v1/pr/report` nowhere. It
+wrote a comment and exited zero, which is the shape of failure this repository
+names most often, with the product itself doing the naming on every pull request
+and nobody reading it.
+
+Both workflows now ask for the credential in their second step rather than
+beside the report at the end, so the check reads as running while a runner is
+working, a job that dies is a run the control plane can name and cancel, and a
+dead run is reported in seconds rather than at the deadline.
--- a/.github/workflows/dogfood.yml
+++ b/.github/workflows/dogfood.yml
@@ -26,9 +26,13 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-# Read, and only read, for everything that runs untrusted code. The one job
-# that needs to write a comment declares it for itself and does nothing else,
-# which is narrower than granting it here and hoping.
+# Read, and only read, for everything that runs untrusted code. The two jobs
+# that need more declare it for themselves and nothing else, which is narrower
+# than granting it here and hoping: the comment job takes
+# `pull-requests: write`, and the job below takes `id-token: write`, which is
+# not a write to this repository at all. It lets the job ask GitHub to sign a
+# statement about itself, which is what it trades for a credential good for
+# this one commit on the control plane.
 permissions:
   contents: read
 
@@ -43,6 +47,21 @@ jobs:
     # apart. They run on the schedule below.
     name: control plane
     runs-on: ubuntu-latest
+
+    # `id-token: write` is the whole reason this check reported nothing for as
+    # long as it has existed. Without it the runner sets neither
+    # ACTIONS_ID_TOKEN_REQUEST_URL nor _TOKEN, the job cannot prove what it is,
+    # and every route to the control plane is closed: the two steps below skip
+    # and the engine's own event sink finds no identity to exchange. The job
+    # then exits zero having posted a comment, and the App, hearing nothing,
+    # says "Nothing was verified" on every pull request this repository has.
+    #
+    # It is not a write to the repository. GitHub mints a short lived signed
+    # statement about this job, and it withholds it from a fork's pull request,
+    # which is the fork case working rather than failing.
+    permissions:
+      contents: read
+      id-token: write
     # Larger than the budgets inside it, deliberately. The harness fails a step
     # that runs over its budget and writes a report saying which one and by how
     # much; a runner timeout kills the job and writes nothing, so a budget that
@@ -69,6 +88,17 @@ jobs:
           --health-retries 10
 
     env:
+      # The hosted control plane this repository is connected to, and the one
+      # whose App posts the `Antifailure` check on every pull request here.
+      #
+      # A literal rather than `vars.AF_CONTROL_PLANE`, which is what
+      # examples/github-workflow.yml tells a customer to set. A customer's
+      # address is their own and belongs in a variable; this repository's is
+      # this repository's own product at its published address, and a variable
+      # that has to be set by hand before the check can work is a variable
+      # somebody will not set. It was not set here, which is the second of the
+      # two reasons nothing was ever reported.
+      AF_CONTROL_PLANE: https://app.antifailure.dev
       # The source the masking rules are written against. Synthesised rather
       # than restored from production: a control plane that copies a real
       # customer's database into a public CI runner would be the single worst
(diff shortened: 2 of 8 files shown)


VirSanghavi:
### Follow-up on the fork case, found by testing it rather than reasoning about it

The claim step I added ran under `set -euo pipefail` and read the two runner
variables bare. GitHub sets **neither** of them on a pull request opened from a
fork, so `set -u` would have ended that step with `unbound variable` and exit 1.

That is precisely the outcome the brief said to avoid: the one case that has to
degrade gracefully would have been the one case that went red, on a check an
outside contributor cannot fix.

Proved, not argued:

```
--- fork case, both unset ---
This job was granted no workflow identity, so it cannot introduce itself.
exit=0
--- old shape, bare read under set -u ---
oldtest.sh: line 2: ACTIONS_ID_TOKEN_REQUEST_TOKEN: unbound variable
exit=1
```

Both variables are read through `${...:-}` now, with a sentence saying what
happened, and every curl pipeline carries `|| true` so a control plane that is
down cannot fail a run either. The observability of a run is not the run.

`TestAJobThatRunsTheCheckCanReportIt` now also fails on a bare read, and I put
one back to watch it fail:

```
--- FAIL: TestAJobThatRunsTheCheckCanReportIt
  job "dogfood" reads ${ACTIONS_ID_TOKEN_REQUEST_TOKEN} bare under set -u, so a
  pull request from a fork, where GitHub sets neither, ends the step with an
  unbound variable and a red check nobody outside this repository can fix.
```

### One thing to expect on this pull request

The `Antifailure` check on **this** pull request will probably still say
"Nothing was verified", and that is not the fix failing.

The workflow half is on this branch and runs here. The control plane half is
deployed, and production is serving v1.1.0 from `main`, which still lets the
first workflow run it sees end the generation. So a stranger's run will close
this commit's generation before or while the claim happens, and the claim will
then be refused with "the check on `a874589` is already unverified", which my
step handles by printing and exiting zero.

The two halves only demonstrate each other once this merges and the control
plane redeploys. The tests are what stand in for that here, and both halves fail
on the tree as it was.

VirSanghavi:
### The check on this pull request reproduced the defect again, on schedule

As said above, and worth recording because it is the same run doing it twice on
two different commits.

Commit `a874589`:

| event | time |
| --- | --- |
| check created | 10:12:37Z |
| `Security` finished, green | 10:14:46Z |
| check completed, "Nothing was verified" | 10:15:31Z |
| `Dogfood`, the job that runs `af ci` | still running at 10:14:49Z |

Two minutes and twenty seven seconds from created to concluded, on a check whose
job needs twenty. Same shape as `ada5644`, same workflow crossing the line
first, and the deployed control plane is still v1.1.0 from `main`, which is
where the guard in this branch is not yet running.

So the claim step could not bind: by the time it asked, the generation was
already `unverified`, and `issueCallback` refuses one that is not `queued` or
`running`. My step handled that the way it is supposed to, printing why and
exiting zero rather than reddening the job.

**This check is expected to stay red on this pull request.** It goes green, or
goes accurately red on a real verdict, on the first pull request after this
merges and the control plane redeploys. The two failing-first test suites are
what stand in for that here.

VirSanghavi:
### The run log, which settles the last two questions

**`af ci` on this repository passes.** The brief asked me to say plainly whether
turning the check on turns a misleading amber into an accurate red. On this
commit it does not: the `control plane` job ran the whole product end to end and
every step succeeded, including `af ci` with the new `--report-json`.

```
success  Tell the control plane this run is the one checking this commit
success  Refresh the golden, with the attestation required
success  The whole check, the way a customer runs it
skipped  Tell the control plane what this commit did
success  Nothing was left behind
```

I tuned nothing to get that. One green run is one green run, not a promise about
the next one, but the honest answer today is green rather than red.

**And the identity works.** The claim step got past minting and was refused by
the control plane, not by GitHub:

```
AF_CONTROL_PLANE: https://app.antifailure.dev
HEAD_SHA: a87458920901d691cf65251cdfd86322cb05558b
The control plane issued no callback credential for a874589...
That is not a failure of this run: it reports through the comment below.
```

`Security` finished green at 10:14:46 and the claim ran at 10:15:08. Twenty two
seconds. `issueCallback` refuses a generation that is not `queued` or `running`,
and by then this one was already `unverified`, so the publish step skipped and
the job stayed green. That is the race in the deployed control plane, caught on
the wire, on the pull request that fixes it.

The `Antifailure` check is therefore red here for the reason documented above,
and every other check on this pull request passes.

PR #221, opened 2026-09-04
npmaudit: a slow registry could exhaust the job without a report

VirSanghavi:
The npm advisory gate could fail without a usable reason or run until GitHub killed it without any report. npm 11 stores registry failures in a root message that the parser discarded. Eight sequential audits could also exhaust the Security job's fifteen minute budget before the tool returned.

Security run 33827299687, job 100882580579, confirms the second failure. Its annotation says `The job has exceeded the maximum execution time of 15m0s`, and cleanup killed the still-running npm audit process. This is separate from the five minute npm refusal in run 33825232140.

The gate now preserves npm's root message, structured error fields, process error and stderr. Empty structured fields are named explicitly. Four workers audit independent lockfiles, with five minutes per project and eleven minutes overall. Each result prints immediately with its project and elapsed time. Stalled processes are terminated, retained output pipes have a two second cleanup limit, and projects still queued at the overall deadline are named as not started. An incomplete scan fails before the policy summary can count it as audited or misclassify an acceptance as stale. Failed output writes also fail the gate.

There is no whole-audit retry. The security policy and required contexts are unchanged.

Verification:

- The compiled gate ran all eight real lockfiles: zero advisories, zero uncovered projects, zero expired or stale decisions. Individual audits took 407 to 831 milliseconds during this run. Earlier API and console audits took 244 and roughly 210 seconds, so a currently responsive registry does not replace the deadline proof.
- Focused race tests passed. Vet passed. Lint reported zero issues. Prosecheck reported zero violations. Gatecheck matched 78 gates across nine workflows.
- Twenty one new isolated mutation cells failed their intended assertions and passed after restoration. They cover overlap, worker bound, collected findings, retained failures, later projects after failure, both progress outcomes, project deadline, overall deadline, queued-project naming, header and result write failures, actual subprocess termination, timeout project, stderr and process error, incomplete and complete entry-point summaries, retained-pipe cleanup and refusal of an incomplete pipe result.
- The ten earlier diagnostic and report-writer mutation cells remain covered.
- A real npm negative control used an unreachable local registry. All eight projects were named inconclusive with elapsed time, ECONNREFUSED, process error and stderr; the gate exited 1 without an audited summary. Removing that temporary registry override made the same binary audit all eight successfully again.

The subprocess tests invoke a real child through the npm entry point, including a child that prints clean JSON but hangs and a descendant that holds an output pipe. A clean-looking partial report cannot make either case pass.


VirSanghavi:
Evidence from this branch's own run, because it failed differently from the ones the description is built on and the difference changes what is left to do.

**This run was killed by the job budget, not by npm's fetch timeout.** Security run 33827299687, job 100882580579, on `4054d614`. The step started at 01:51:15 and the runner cleaned up at 02:06:09, which is 14m54s against `timeout-minutes: 15` in `.github/workflows/security.yml:97`. The last lines are:

```
Terminate orphan process: pid (2375) (go)
Terminate orphan process: pid (2410) (npmaudit)
Terminate orphan process: pid (2608) (npm audit)
```

`npm audit` was still running when the budget expired. There is no `npmaudit:` line anywhere in the log, so the diagnostics this branch adds never got the chance to fire.

**So there are two distinct failure modes and this branch addresses one.**

| Run | Duration | What ended it |
| --- | --- | --- |
| 33815005859 | 5m01s | npm's 300s fetch timeout, empty refusal |
| 33825232140 | 5m01s | same |
| 33827299687 | 14m54s | the job's 15 minute budget, nothing reported |

**Why the budget is the binding constraint.** `run` at `tools/npmaudit/main.go:165-172` loops over the locked projects sequentially and returns on the first error, and there are eight lockfiles tracked: `api`, `console`, `docs`, `ee/web`, `examples/next-app`, `runner`, `web`, `www`. The measurements in this PR's description, 244s for the API and roughly 210s for the console, are already 7.5 minutes for two of them. Eight projects at anything like that rate cannot finish inside fifteen minutes, and when they do not, the gate reports nothing at all rather than reporting late.

Nothing above argues against this change. The decision not to add a whole-audit retry looks right and this run supports it: the budget, not npm's retry policy, is what ran out. What it does mean is that the diagnostics alone will not turn the job green, and a run that dies at the budget is invisible to them.

Worth considering as follow-up, not asked for here: a per-project deadline shorter than the job budget so the tool always gets to say which project it was on and how long it waited, and auditing the projects concurrently, since eight independent registry calls have no reason to be serialized.

I had opened #222 for the same defect before seeing this. Closed as a duplicate; this one is better, and specifically it decodes npm's root `message`, which #222 did not, so #222 would still have printed an empty refusal in exactly the case that took main red.

PR #222, opened 2026-09-04
A refusal that named no cause

VirSanghavi:
`npm advisories` is red on main and on every open pull request. The entire output is:

```
npmaudit: api: npm audit refused:
exit status 1
```

Nothing after the colon. npm returned a report whose `error` object was present and whose code, summary and detail were all empty, and the format string at `tools/npmaudit/main.go:324` printed the three empty strings faithfully.

## What the timing says

The step ran for exactly five minutes before failing, in both failing runs:

| Run | Started | Failed | Duration |
| --- | --- | --- | --- |
| 33815005859 | 22:51:54 | 22:56:55 | 5m01s |
| 33825232140 | 01:18:20 | 01:23:21 | 5m01s |

That is npm's own default `fetch-timeout` of 300000ms, and it is the shape of a registry that never answered rather than of a real advisory. None of that evidence reached anybody.

## Why this one path

`parse` has three error paths. The two that fire on a malformed report both call `trim(stderr)`. This one, the only one that fires when the registry is unreachable, was the one that threw the diagnostics away. So the gate that exists to say what is wrong with the dependency tree could not say what was wrong with itself, and there is no way to tell a real advisory from an unreachable registry without reading the source.

## What changed

A refusal now carries npm's exit error and its stderr. An empty refusal says it is empty rather than trailing off after a colon. An empty stderr is stated out loud, because npm printing nothing and this tool discarding what npm printed look identical in a log, and the second of those is the defect being fixed.

**This does not by itself make the gate pass.** It makes the next failure legible, which is the prerequisite for fixing the cause rather than guessing at it. If the registry is the problem, the next run will say so in npm's own words.

## Verification

Every test was mutation tested, one break per assertion:

| Mutation | Caught by |
| --- | --- |
| Restore the original format string, the exact bug | `TestAnEmptyRefusalSaysThatItIsEmpty` and two others |
| Drop stderr from the refusal | `TestARefusalCarriesStderrAndTheExitError`, `TestAnEmptyStderrIsStatedRatherThanOmitted` |
| Drop the exit error | `TestARefusalCarriesStderrAndTheExitError` |
| Treat every refusal as empty, dropping populated fields | `TestAPopulatedRefusalStillNamesItsCause`, `TestNpmRefusingIsAnErrorAndNotAnEmptyReport` |
| Keep the trailing colon on an empty refusal | `TestAnEmptyRefusalSaysThatItIsEmpty` |

The pre-existing `TestNpmRefusingIsAnErrorAndNotAnEmptyReport` still passes, and a new test asserts a populated refusal is not reported as empty, so the fix to the blank case cannot bury the case that already worked.

### DIFF
--- a/.changes/a-refusal-that-named-no-cause.md
+++ b/.changes/a-refusal-that-named-no-cause.md
@@ -1,0 +1,19 @@
+# fixed
+
+The npm advisory gate refused and would not say why.
+
+`npm advisories` went red on main and on every open pull request, and the whole
+output was `npmaudit: api: npm audit refused:` with nothing after the colon. npm
+had returned a report whose `error` object was present and whose code, summary
+and detail were all empty, and the format string printed the three empty strings
+faithfully. The step had run for exactly five minutes first, twice, which is
+npm's own default fetch timeout and the shape of a registry that never answered.
+None of that reached anybody, because this is the one error path that fires when
+the registry is unreachable and it was the one path of the three that did not
+carry stderr.
+
+A refusal now carries npm's exit error and its stderr, and an empty refusal
+reports itself as empty rather than trailing off after a colon. An empty stderr
+is stated out loud, because npm printing nothing and this tool discarding what
+npm printed look identical in a log, and the second of those is what was
+happening.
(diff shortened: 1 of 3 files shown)


VirSanghavi:
Closing as a duplicate of #221, which is the better change. Both preserve the process error and stderr and both say so when the structured fields are empty. #221 also decodes npm's root `message`, which is where npm 11 actually writes the network failure, so this branch would still have printed an empty refusal in the exact case that took main red. It also carries a real local reproduction rather than an inference from timing: the API audit completes in 244 seconds with zero advisories and a controlled unreachable-registry run produces the refusal shape. I opened this without checking the open pull request list first, which is on me.

PR #264, opened 2026-09-05
deploy: the careers smoke had no browser to drive

VirSanghavi:
The new post-publish step went red on the first push to main that ran it. It
installed the runner with `npm --prefix runner ci` and then installed the
browser with `npx playwright install` from the repository ROOT.

There is no package.json at the root. So npx did not resolve the runner's
pinned `^1.49.0` at all: it fetched the newest playwright from the registry and
downloaded browsers for that one. The runner then launched its own version and
looked for a revision nothing had installed, `chromium_headless_shell-1234`.

Both commands now run inside `runner`, which is what dogfood.yml has done since
it was written.

WHY THIS WAS NOT CAUGHT BEFORE IT MERGED, which is the more useful half.
`deploy.yml` runs on push and not on pull_request, so the pull request that
added the step never ran it. The step was wired, reviewed, merged, and had
never once executed. That is the same shape as any other dead path: everything
about it looked right and nothing had asked it to work.

WHAT IT COST, AND WHAT IT DID NOT. The smoke reported COULD NOT TELL and exited
2 rather than reporting a pass, so this was a red rather than a false green. Its
own words: a smoke that cannot say "I did not find out" says "fine" instead.
That is the behaviour it was built for and it is the reason this is a small
commit rather than an incident.

Proved by running the whole step locally against the live site rather than by
reading it:

    ok        https://antifailure.dev      the page showed what it shows when
                                           this works
    REFUSED   https://www.antifailure.dev  The page shows an error rather than
                                           what was expected. It says: "Could
                                           not reach the server."

The second line is the www hostname bug, which is fixed on main and not yet in
production because production moves only on a tag. It is not this commit's to
fix and the smoke is right to keep saying so.


VirSanghavi:
Superseded by a clean branch: this one accidentally carried ten evidence files from a local smoke run, which TestEveryTopLevelPathIsClassified correctly refused.

PR #277, opened 2026-09-06
main: the merge commit of #276 carries no sign-off, so cd deployed nothing

VirSanghavi:
#276 was squash merged with gh pr merge and a body file, so the commit on
main carries no Signed-off-by. That is the failure tools/prmerge was written
for and CONTRIBUTING.md names the remedy, just merge, which the agent did not
use. The DCO job is a required context, it went red on b5fbcbb7, and cd.yml's
gate read that and deployed nothing, so staging does not have migration 0040
or any of #276.

This commit changes no file. It exists so that the head of main carries a
sign-off again and cd deploys it, and it is merged with just merge.


VirSanghavi:
Unnecessary now: main went green on 4ef50aac with the required contexts passing, so cd deploys from there and this signed empty commit has nothing left to do.

PR #278, opened 2026-09-07
engine: a conformance suite that had never said no, and the two checks that could not

VirSanghavi:
## A control that could not fail, in the instrument built to prove things fail

The most useful thing this lane produced is a defect in its own work.
`fakes.InMemoryDatabase` derived a branch's `ProviderRef` from the environment
identifier, so two creations for one environment agreed whatever the code did.
The assertion that they agree therefore stayed green with the idempotence check
deleted. It was written, read, believed, and worthless, and only mutation
testing said so. The reference counts up now, and deleting the check goes red.

That is the same shape as the two defects the lane was built to find.

## Two checks that could not say no

`engine/conformance/db.go` admitted in its own package doc that the database
conformance suite had never been shown to fail. Twenty four behaviors were
declared, four providers ran them, every run printed ok, and nobody had ever
watched one go red. Making it go red found two assertions that were not
checking what they claimed.

**`Cancellation_LeavesNoUntrackedResource` had its polarity inverted, in both
directions at once.** It failed when the call reported an error, returned a
provider reference, and the inventory listed that reference. That is a provider
doing exactly the right thing: telling you what it made so teardown can remove
it. The Docker provider has such a path, in `waitReady`, which returns a
populated branch alongside an error. Meanwhile the case the rule is actually
about, a resource created and NOT named in the result, was never looked for,
because the loop only ran when the result named something. The message said
"that the caller has no identifier for" while the code required the caller to
have one. It passed everywhere because no shipped provider creates anything on
a cancelled call, which is the same reason nobody noticed.

**`Branch_RefusesAnUnverifiedGolden` could not reach the branch side of its own
rule.** That was recorded in `knownGaps` as needing an interface change to
close. It did not need one. A provider that FLAGS an unverified version rather
than withholding it is something the suite already permits:
`Refresh_RefusesToPublishWhenVerificationFails` asserts on the `Verified` flag,
not on the absence of a version. `PublishUnverified` on the fakes is that
affordance, it is not a fault, and
`TestTheUnverifiedAffordanceBreaksNothingOnItsOwn` runs the suite green with it
alone, so the red under `branch-accepts-unverified` is attributable to the
fault and not to the premise. `knownGaps` is now empty, and kept, because the
next gap needs somewhere to be recorded.

## The number

**24 of 24 behaviors proved able to fail, each with the break that proves it.**

    cd engine && go test ./conformance/... -count=1 \
        -run TestEveryBehaviorIsProvedAbleToFail -v

Apple M3, 16 GB, macOS 26.3.1, go1.27.0 darwin/arm64, run under the shared
build lock this repository serialises Go on, against PostgreSQL 17.11 in the
`postgres:17-alpine` container on 55432 that `just db` and CI both start. The
test prints the twenty four row table itself, with the behavior, the break,
which provider proved it, and the assertion that went red, so the count is
checkable by somebody who does not trust the count.

## How it is proved

A previous pass built `fakes.InMemoryDatabase` and `fakes.Break` and covered
nine behaviors. The other fifteen were out of reach for two different reasons.

- **Five read rows back through a connection string**, and a fake with no bytes
  can only agree with whatever it was told, which is the same thing as not
  checking. `fakes.NewPostgresDatabase` is the second provider: each golden and
  each branch is its own database on a real Postgres, a branch is
  `CREATE DATABASE ... TEMPLATE`, and the inventory reads `pg_database` rather
  than the provider's own memory, because a leak detector that trusted the
  provider would report exactly the resources the provider remembered to
  mention. Isolation is then a property of Postgres rather than of the fake.
- **Ten simply had no fault written.** There are sixteen new ones, plus a
  second fault for two behaviors whose two rules are caught by two different
  clauses. A clause no break exercises is a claim nobody has checked.
- The in-memory fake now declares every capability it can honestly support. An
  undeclared capability is a SKIPPED behavior, and a self test whose subject
  skipped is the false green the whole file is against.

## The checks that can say no

`requireRan` fails three separate ways, because two of them exit zero and read
exactly like a pass: a behavior that failed, a behavior that SKIPPED, and a
`-test.run` that matched no test at all. `runChild` refuses a child whose fault
could not be injected, so "not injected" can never be recorded as "injected and
survived"; `Break` returns `fakes.Uninjectable`, whose every method fails
loudly, rather than a quietly ordinary provider. `DropEverything` asks
`pg_database` again after its own loop rather than reporting success because it
ran.

Without a Postgres the five row behaviors are named in the table as NOT PROVED
HERE with the address that did not answer, and the count says 19 of 24 rather
than 24 of 24. CI sets `AF_REQUIRE_DATABASE`, which turns the missing server
into a failure before the table is reached, so the gate cannot pass that way.
Both of those were exercised by pointing `AF_TEST_DATABASE_URL` at a closed
port.

An exit code is not a verdict, and that came up three times while this was
being built: a broad sweep reported exit 0 after being killed mid compile
having reached no test at all, a run whose output had not flushed looked
finished and empty for minutes, and a `-test.run` that matches nothing exits
zero and prints ok. Only the third is inside the test, and it is the one this
PR builds a check for.

## What was NOT checked

- `./conformance/...` under `-race`. `./internal/testutil/fakes/...` passed
  under `-race` in 27.578s; the conformance package was stopped mid build with
  790 MB of swap left on a 16 GB machine running four lanes, because a race
  build is the heaviest thing on it. CI runs the whole engine with `-race` on a
  clean runner, which is where that answer comes from.
- The broad local sweep of `./chaos/...` and `./internal/db/...`. It was
  stopped after roughly 45 minutes for the same reason, having built no test
  binary. The Docker provider's `Cancellation_LeavesNoUntrackedResource` was
  run against the real daemon on its own and passed in 51.27s, which is the
  behavior this PR changes; the rest of that sweep is CI's.
- `neon`, `supabase` and `dblab` conformance, which skip by name without
  credentials, as they are designed to.

`.changes/a-suite-that-had-never-said-no.md`. `docs/plan/STATUS.md` row 1.9 and
one line of `docs/plan/QUESTIONS.md` are updated to what is now true.


VirSanghavi:
A pattern worth naming, because this landed four instances of it in one night
and the last of them is what this PR builds a check for.

An exit code says how a process ended, never whether the work happened.

1. A broad `go test ./chaos/... ./internal/db/...` sweep, killed after 45
   minutes having built no test binary, reported exit 0 through the
   backgrounded shell.
2. A finished `-race` run whose output had not flushed through a pipe looked
   finished and empty for several minutes. Its real line was
   `ok .../internal/testutil/fakes 27.578s`.
3. A duplicate run killed with SIGTERM reported exit 143, which reads exactly
   like a test failure.
4. `go test -run <pattern>` matching nothing exits zero and prints ok.

Only the fourth is inside a test, and `requireRan` in `db_selftest_test.go` now
refuses it by requiring the child's output to carry a PASS line naming that
exact subtest, rather than trusting the child's status. The other three are the same defect
arriving through the process table: a verdict that cannot distinguish "passed"
from "never ran".

The general form is the one this repository keeps rediscovering one instrument
at a time, and it is already written down for CI in CLAUDE.md, five things
render as a red check and only one means a test failed. It is equally true of
green, and equally true locally.

One more, found while sweeping up: mutating `DropEverything` so it dropped
nothing left three databases on the shared cluster. The cell passed and the
leak it proved was real was also real. A mutation runner restores the source;
it does not undo what the broken code did, and the cells most worth writing are
the ones that break cleanup.

PR #310, opened 2026-09-08
ee: the resolver rule group that blocked nothing, and two paths called open for the weaker half of the question

VirSanghavi:
Draft opened centrally so CI runs. `ci.yml` fires on pull_request and on push to main only, so a pushed branch with no pull request gets no verdict at all.

The lane owns this branch and will replace this description with the failure it fixes, its acceptance evidence and its mutation table before it is marked ready.

Opened by the merge coordinator, not by the lane.

### DIFF
--- a/.changes/a-dns-rule-group-that-blocked-nothing.md
+++ b/.changes/a-dns-rule-group-that-blocked-nothing.md
@@ -1,0 +1,41 @@
+# security
+
+A Route 53 Resolver DNS Firewall rule group that let every query through was
+reported as closing the resolver.
+
+AWS processes a rule group by order of priority starting from the lowest, and
+defines ALLOW as permitting the request to go through and ALERT as permitting it
+and sending metrics and logs. So a group holding ALLOW on every domain at
+priority 5 and BLOCK on every domain at priority 1000 blocks nothing at all, and
+the check called it closed because every assertion it made about the terminal
+rule was true. It reads as configured in a console screenshot. The check now
+refuses that shape, a terminal rule moved off the end by priority, two rules
+sharing the last priority, which AWS refuses to create, and a domain with a star
+anywhere but the front, which a DNS Firewall domain list cannot hold. Each has a
+mutation aimed at it.
+
+The failure mode of that association was a bool and the AWS field takes three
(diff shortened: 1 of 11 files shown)


VirSanghavi:
Closing this as **superseded by #335**, where **this branch's ECS package is the
base**, unchanged.

**Why it is superseded rather than merged.** #313 `w-runtime-ecs` created
`ee/engine/runtime/ecs/` from a base where it did not exist, and so did this one.
They are not two parallel pull requests: this branch is #313's first two commits
carried forward under different shas and then continued for six more. The ECS
package is byte identical at the shared point. **Merging both in sequence would
have added the package twice**, and neither git nor the Go compiler would have
said so.

#335 is the union off current `main`. This branch's eight commits are cherry
picked onto it and the result is byte identical to this one:

```
git diff origin/w-ecs-close-egress <335 head> -- ee/engine/runtime/ecs/
# empty apart from one assertion ported from #313
```

**Everything this branch found is preserved**: the resolver rule group that
blocked nothing, the three valued `FirewallFailOpen` with its caller still
assigning true, the essential probe container that would have killed the task it
was probing for, the two paths that were called open for the weaker half of the
question, the licence headers, the evidence recording and its grades, and the two
mutations proving a check a different check was catching first.

**What #313 added on top, since it is the half that is easy to lose:** the
enterprise runtimes documentation page, the `kubelet` dictionary entry, the
regenerated `pages.gen.go`, and one assertion. The page is rewritten rather than
copied, because #313's version publishes eight of thirteen paths closed and this
branch closed two more, so #335's page says **ten closed, one open, two
unproven**, which is what the binary actually prints.

**Two changes to this branch's own material in #335, both small and both
explained there.** The changelog entry said the report "went from eight of
thirteen paths closed to ten", which is true of the branch and false of the
release, since no version ever carried eight; it says what the ten are instead.
And the time sync verdict's detail ended "A probe from one running task settles
it and none has been recorded for this environment", which reads as a measurement
pending when this path declares no `Observe` hook and `Targets()` names no probe
for it, so there is nowhere to put an answer. The design is right and unchanged.
Only the sentence moved.

Nothing is lost. The branch `w-ecs-close-egress` is left in place for the record.

PR #313, opened 2026-09-08
ee: a manifest naming the ecs runtime was answered by a list of two names

VirSanghavi:
## What this is

Lane L5.1, AWS ECS on Fargate. It does not add a runtime. It adds the
enumeration of every way out of an ECS task, a predicate over the configuration
such a runtime would generate, and a refusal that prints both.

Section 10 of the plan carries a standing risk: the containment claim must not
soften to make a cloud runtime possible, and if ECS cannot be proved contained
then the honest answer is that Antifailure runs on EKS and not on raw ECS. This
is that answer, with the work behind it rather than as an opinion.

## The number

**13 distinct egress paths out of a Fargate task. 8 closed by the generated
configuration, 3 open, 2 unproven.**

Produced by `TestTheNumber` in `ee/engine/runtime/ecs`, and printed by the
binary itself when a manifest names the runtime.

The three open paths are open because closing them removes the environment: the
ECR and CloudWatch Logs interface endpoints and the S3 gateway endpoint the
image pull needs, plus the task metadata endpoint, which AWS documents as on by
default with no way to turn it off. The two unproven ones are named rather than
rounded up.

## Two lines of the lane row are wrong

**"a security group that denies egress" cannot exist on Fargate.** On platform
version 1.4.0 the ECR login, the image pull and the log push all flow over the
task ENI under the task's own security group, and AWS states that a Fargate task
must have a route to the registry to pull an image. There is no kubelet here
pulling outside the pod's policy. Denying egress produces a task that never
starts, not a task that cannot escape.

**"IMDS disabled for the task" is not a field.** The documented defence is
`ECS_AWSVPC_BLOCK_IMDS`, an ECS container agent variable set on an EC2 container
instance, and no task definition can set it. On Fargate no EC2 instance profile
exists for the endpoint to vend, but AWS nowhere states that 169.254.169.254
stops answering, and those are different claims. The verdict is therefore
`unproven`, not `closed`.

## The finding worth carrying to L5.2 and L5.3

On AWS the security group is irrelevant to a DNS query. The VPC user guide states
that traffic to and from the Amazon DNS server cannot be filtered with network
ACLs or security groups. The Route 53 Resolver answers recursive queries for
public names from anywhere in the VPC, and a DNS question's payload is chosen by
whoever asks it. Only a Route 53 Resolver DNS Firewall rule group whose last rule
blocks every domain, and which does not fail open, closes it.

A containment argument carried over from Kubernetes gets this wrong, because
there the same attempt is closed by the NetworkPolicy that closes everything
else. Cloud Run and Container Apps each have their own version of this and each
lane owes its own primary source.

## What is proved and what is only asserted

Proved: the predicate. Every path reported closed carries a mutation that breaks
the mechanism it depends on and requires the verdict to change, and a separate
test refuses any closed path with no mutation aimed at it.

Asserted, not proved: the eight closed verdicts themselves. They are statements
about a JSON document that has never been applied to an AWS account. No packet
has been observed failing to leave a task. `Report.Caveat` is a constant so this
sentence travels with every number the package produces and cannot be
paraphrased away.

`conformance.RunRuntime` is NOT green here. It cannot be, because there is no
runtime for it to run against. Nothing in this branch claims otherwise.

## The conformance run this proposes and does not perform

One run, in a dedicated AWS account used for nothing else, to settle the two
unproven paths and to check that AWS enforces what the plan says. One Fargate
task in one throwaway VPC, running the escape script and exiting. Ceiling: 25 US
dollars. Teardown proved by a `resourcegroupstaggingapi` inventory of the run's
tag returning empty, recorded in the run report. Not performed tonight, and it
needs Vir's explicit approval because it spends money under his account.

## Section 2.4 item 6: three lines of the plan found wrong

Two are in the lane row and are stated above. The third is the row's edition.

**The row places ECS in `/ee`, and nothing in this repository justifies that
today.** Section 2.5 says a runtime is MIT unless it carries licensed
governance, and `local` and `kubernetes` are both MIT. The only candidate
feature that exists is `multi_runtime`, and it does not fit twice over: it has
**zero enforcement sites** anywhere in the tree, appearing only in its own
declaration, in `tools/licensegen`, in a test asserting it is off, and in the
docs; and `docs/enterprise/runtimes.md` defines it as placement across several
runtimes rather than as the right to have another one, so gating a runtime on it
would make that page wrong. The `cloud_runtime` feature that would fit is
L6.2's row and does not exist yet.

There is a stronger argument than the letter of 2.5 for moving it, and it is
about what this package actually does. **This is not a runtime, it is the
product telling the truth about where it cannot run.** Behind the enterprise
licence, a community user who points a manifest at ECS gets no refusal at all
and is left believing it works. The refusal is worth more to the person who has
not paid than to the person who has.

The package stays under `ee/` in this branch on the team lead's instruction, who
is sequencing that move across this lane and L5.3 at merge time rather than
having each lane do it independently.

## A shared gate finding for the next `/ee` package

`tools/licensecheck` is a required context and it requires the literal string
`Antifailure Enterprise License` in the **first 400 bytes** of every `.go`,
`.ts` and `.tsx` file under `ee/`. A new package under `ee/` fails it on every
file it adds. This lane hit it on 4 of 4 new files and L5.3 hit it on 4 of 4 of
its own, independently, which makes it a property of adding a directory rather
than anybody's oversight.

### DIFF
--- a/.changes/ecs-runtime-containment-report.md
+++ b/.changes/ecs-runtime-containment-report.md
@@ -1,0 +1,20 @@
+# added
+
+`runtime.provider: ecs` is now answered by a containment report instead of by a
+list of the two runtimes that exist. It still refuses to start an environment,
+and the refusal is the feature: it enumerates thirteen distinct ways out of an
+ECS task on Fargate, names the eight the generated configuration closes and the
+five it does not, and says which verdicts are computed from a configuration
+nobody has applied to an AWS account.
+
+Two things the enumeration found are worth knowing even if you never run on ECS.
+A security group cannot deny egress on Fargate, because the image pull runs over
+the task's own interface under that same security group, so denying egress
+produces a task that never starts. And on AWS a security group cannot filter a
+DNS query at all: the Amazon provided resolver answers recursive lookups for
+public names from anywhere in the VPC and the VPC user guide states plainly that
+neither security groups nor network ACLs can filter traffic to it, so closing
+that path takes a Route 53 Resolver DNS Firewall rule that blocks every domain.
+
+The honest summary is unchanged and is now written down: Antifailure runs on EKS
+and not on raw ECS.
--- a/docs/src/content/docs/enterprise/runtimes.md
+++ b/docs/src/content/docs/enterprise/runtimes.md
@@ -50,4 +50,81 @@ What the enterprise edition adds here is not a third runtime. It is placement
 across several of them at once: the requirements, the tags and the scheduling
 described above.
 
+## Why there is no ECS runtime
+
+`runtime.provider: ecs` is registered in the enterprise binary and it refuses,
+every time, with a report rather than an error. The reason is worth reading
+before assuming it is a gap somebody will close next release.
+
+A runtime is allowed to exist here only if it can prove an environment has no
+way out, and the Kubernetes runtime proves it the only way a proof works: it
+creates the `NetworkPolicy` objects itself, then runs one pod under exactly the
+rules a service runs under and has it try to escape before any application image
+starts. If any attempt gets out the environment does not start and you get
+**AF-RUN-043**. That is not caution. Several container network plugins accept a
+`NetworkPolicy` object and enforce nothing, every status reads green, and the
+only thing that can tell the two apart is a packet.
+
+ECS on Fargate cannot be given the same treatment, for two reasons that are
+properties of the platform rather than of this implementation.
+
+**The image pull runs inside the boundary.** A kubelet pulls on the node, so a
+pod can be denied every egress rule and still start. On Fargate platform version
+1.4.0 the ECR login, the image pull and the log push all flow over the task's own
+network interface, under the task's own security group, and AWS states that a
+Fargate task must have a route to the registry to pull an image. So a security
+group that denies egress does not produce a contained environment. It produces a
+task that never starts, and the endpoints that let it start are themselves
+reachable addresses.
+
+**The enforcement cannot be observed without an account.** Whether a cluster
+enforces a `NetworkPolicy` is answerable on a laptop in ninety seconds. Whether
+AWS enforces a route table is a fact about an account, and no test in this
+repository is permitted to need one.
+
+So what ships is the enumeration and the predicate over it, and the refusal
+carries both. Thirteen distinct egress paths out of a Fargate task, of which
+**eight are closed by the configuration this runtime would generate, three are
+open, and two are not decided by either the configuration or AWS's own
+documentation**. Every closed verdict is a statement about a JSON document and
+not about an observed packet, and the report says so in those words.
+
+The three that are open are open because closing them would stop the environment
+existing, or because AWS provides no way to close them: the ECR and CloudWatch
+Logs interface endpoints and the S3 gateway endpoint that the image pull needs,
+and the task metadata endpoint, which is on by default for every Fargate task on
+platform version 1.4.0 or later with no documented way to turn it off.
+
+The two that are unproven are named rather than rounded up. The instance
+metadata service at `169.254.169.254` is the sharper of them: the Fargate launch
+type removes the documented credential source, because EC2 instance profiles are
+not available to containers in Fargate tasks, but AWS does not state that the
+address stops answering, and those are different claims. Settling it needs one
+request from one running task.
+
+One finding is worth carrying away even if you never run on ECS. **On AWS the
+security group is irrelevant to a DNS query.** The VPC user guide states that
+traffic to and from the Amazon DNS server cannot be filtered with network ACLs
+or security groups, and that resolver answers recursive queries for public names
+from anywhere in the VPC. A DNS question is chosen by whoever asks it, so that
+is a data channel out that no security group audit will ever show you. Closing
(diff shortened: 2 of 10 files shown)


VirSanghavi:
Closing this as **superseded by #335**, which carries its work forward.

**Why it is superseded rather than merged.** #310 `w-ecs-close-egress` is not a
parallel pull request, it is this branch continued. Both start with the same two
commits, `ee: the ECS containment claim was a sentence nobody had checked against
AWS` and `ee: a manifest naming the ecs runtime was answered by a list of two
names`, under different shas, and the ECS package is byte identical at that
point:

```
for f in egress.go egress_test.go plan.go runtime.go; do
  diff <(git show ddb3fc3b:ee/engine/runtime/ecs/$f) \
       <(git show 0ae1a1d0:ee/engine/runtime/ecs/$f)
done          # no output
```

From there #310 rewrote the package over six more commits. So **merging both in
sequence would have added the package twice**: two
`extension.Default.AddRuntimeProvider(ecs.NewProvider())` calls, two
`engine/api/packages.txt` lines for one import path with two different reasons,
and a documentation page describing a report the binary no longer prints. Git
would not have objected and neither would the Go compiler.

**In #335, #310's ECS package is the base and this branch's contribution is the
documentation.** There is no code here to rescue, because #310 is a strict
continuation.

**The page had to be rewritten rather than copied.** It publishes eight of
thirteen egress paths closed, three open and two unproven. The shipped report
says **ten closed, one open, two unproven**, because #310 closed the interface
endpoints and the S3 gateway endpoint with a VPC endpoint policy. A page saying
eight would be a measurement that was true of a branch nobody merged. Everything
else the page argues is carried over intact, and three things are added: why
those two paths moved, that a DNS Firewall rule group is read from the lowest
priority up, and that the allow list is generated from the manifest's egress
catalogue.

**Also carried forward from here:** the `kubelet` dictionary entry, the
regenerated `pages.gen.go`, and one assertion. This branch's
`TestTheReportCarriesItsCaveat` reads the report's FIRST LINE rather than the
whole output, so the ratio cannot be quoted without the open and unproven counts
arriving with it. That was the one assertion here about the SHAPE of the report
rather than about its numbers, so it carried over unchanged in intent and changed
in figures, and it has three mutation cells behind it in #335.

**One thing here is a bug that #310 fixed and it is worth naming.** This branch
marks the containment probe container `Essential` and gives it a `DependsOn`. On
ECS the exit of an essential container stops the whole task, and the probe is a
thing that runs, prints what it saw and finishes, so this would have **killed
every environment the moment its own containment probe succeeded**. #310 removed
it. That is why the per file decision was made by reading both sides rather than
by taking the newer one, even though the newer one turned out to be right.

Nothing is lost. The branch `w-runtime-ecs` is left in place for the record.

VirSanghavi:
Closing note, added after the fact: **this was closed because its work is on
main, not because a merge was refused.** #310 carries an equivalent note and this
one did not, which is the whole reason for adding it. Two pull requests closed
three seconds apart with no comment on either reads exactly like the
`gh pr merge --delete-branch` failure this repository has written down, where a
refused merge still deletes the branch and closing the pull request is the side
effect. That is not what happened here, and the branches surviving on the remote
is the evidence: that flag would have deleted them.

**Where the work went.** #335 `88627d7f`, merged 2026-09-08T22:35:31Z.

This branch and #310 both created `ee/engine/runtime/ecs/` from a base where it
did not exist. They are not parallel: #310 is this branch's first two commits
carried forward and then continued for six more. Verified rather than read off
the subject lines:

```
git show 37ea05e0 | git patch-id --stable
git show 53aad436 | git patch-id --stable   # same id
git show ddb3fc3b | git patch-id --stable
git show 0ae1a1d0 | git patch-id --stable   # same id
```

Merging both in sequence would have added the package twice, registered the
runtime twice, and written two `packages.txt` lines for one import path. Neither
git nor the Go compiler reports any of that.

**What this branch had that #310 lacked, and where each piece is now.** These are
the half that was easy to lose, so they are listed with what was checked on
`origin/main` rather than asserted:

- `docs/src/content/docs/enterprise/runtimes.md`: on main, **rewritten**.
- `engine/internal/docs/pages.gen.go`: on main, regenerated by its generator
  rather than hand edited.
- `tools/docs/dictionary.txt`: on main, the `kubelet` entry, once.
- `.changes/ecs-runtime-containment-report.md`: on main.

The page is rewritten rather than copied and that distinction matters. **This
branch's page publishes eight of thirteen paths closed, three open and two
unproven.** #310 closed two more, so the shipped report says **ten closed, one
open, two unproven**, which `TestTheNumber` on main now asserts. Copying the page
across would have published a number the binary does not print.

`engine/api/packages.txt` carries exactly one line for the package, in #310's
wording. `ee/engine/cmd/af/main.go` carries exactly one `ecs.NewProvider()`
registration.

**The substance this lane owed is intact and was not softened to let the work
land.** The enumeration is thirteen paths, the predicate refuses, `Contained()`
is false by construction, and the page still ends on the honest answer that
Antifailure runs on EKS and not on raw ECS.

The branch `w-runtime-ecs` is untouched at `ec1c3578` if anything here needs
reading again.

PR #407, opened 2026-09-13
deploy: a declared retry stance read as undeclared, and main went red on a race

VirSanghavi:
## The failure

Main is red on d608a4dc (engine job 103669779416) and 2d0d0dde, in deploy_test.sh:

```
no declared retry stance for: job start
FAIL  every Azure call that mutates state has a declared retry stance
```

2d0d0dde named `job update` instead. Neither stance is missing. Both have been declared in the header of deploy.sh since #280.

## Why

The check ran `stance_text | grep -Fq "$tail"` under `set -euo pipefail`. `grep -q` exits at its first match, and GNU sed writes to a pipe in 4096 byte chunks. #384 added a paragraph that took the header from 4081 bytes to 4381. So sed now makes a second write after grep has gone, dies of SIGPIPE (141), and pipefail reports a declared operation as undeclared. Every operation is first mentioned before byte 3200, so any of them can lose.

#384's own engine job tested "Merge 4ede3366 into 957eb0c3", the same header, and won the race. macOS buffers pipes differently and passes every time.

## Evidence, Linux (python:3.12-slim, GNU sed 4.9, bash 5.2, no network)

- Before #384 (957eb0c3): the check's functions refused 0 of 300 runs, PIPESTATUS [0 0] every time.
- Unmodified main: 300 of 300 runs refused at least one operation. The bare pipeline returned [141 0] in 272 of 300. The full deploy_test.sh exited 1 in 5 of 5 runs, each naming a different set.
- This branch: 0 of 300 runs refuse, and 5 of 5 full runs pass with 78 assertions. macOS passes too.
- Still able to say no: renaming `job start` in the header refuses naming only `job start`, and renaming `traffic set` refuses naming only `traffic set`. Each fails only this assertion.

## The change

The check reads the header once and matches with a shell pattern, with no pipe to cut short. The two other early exiting pipelines in the file do not have this race. Line 569 uses `grep -c`, which reads everything. `line_of` pipes into `head -1`, but its largest output across a full run is 70 bytes, one write.

This does not change deploy.sh or any stance. It must merge before anything else, because every PR's engine job runs this test against main's header.


VirSanghavi:
Superseded by #406, which fixes the same pipefail SIGPIPE race in deploy_test.sh and was opened first. This PR's Linux evidence (full deploy_test.sh red 5 of 5 on main; mutation cells renaming `job start` and `traffic set` refuse only that operation) is carried into #406's merge message.

PR #412, opened 2026-09-13
throwaway: reproduce the dogfood oracle refusal on an old base

VirSanghavi:
A throwaway draft, not for merging. It will be closed and its branch deleted.

The branch is cut at 51b9ed9c, an older main, and carries only an empty commit. Its purpose is to reproduce, on the pull request shape that failed on #374, the dogfood job refusing its oracle baseline with AF-ORC-003 and then failing its scenarios with AF-LOD-010, and then to show the workflow fix passing on the same shape.


VirSanghavi:
Throwaway proof finished: the dogfood oracle refusal reproduced on a stale recorded base, and passed with the fix on the same shape. Closing and deleting both throwaway branches.

PR #419, opened 2026-09-13
ee: an EC2 instance role could list itself and never be read

VirSanghavi:
## The failure

Every EC2 instance role was unreadable. The AWS credential chain in `ee/engine/cloudauth` asks instance metadata for an IMDSv2 session token and sends it when listing the role. Then it reads the role's credentials through `fromEndpoint`, which was written for the ECS endpoint and took an Authorization token. Instance metadata passed it an empty one, so the credentials request carried no `X-aws-ec2-metadata-token`. An instance that requires version 2, the only kind this chain reads, answers that request 401.

`discover` then dropped the error and said instance metadata did not answer.

## Who reads credentials through this chain

Every place `cloudauth.NewAWSChain` is built, from `git grep` on main with tests excluded, and cross-checked by lane-rds:

- `ee/engine/db/aurora/aurora.go:283`, the Aurora provider. Its control plane client in `db/aurora/rds.go` signs with this chain.
- `ee/engine/secrets/aws.go:101`, the AWS Secrets Manager secret source (`AWSBackend`).
- `ee/engine/db/rds/rds.go`, the RDS for PostgreSQL provider on #400. That branch is not merged, so it inherits this fix when it rebases.

On an instance role, none of them could start. `docs/enterprise/secrets.md` and the Aurora provider page both promise the EC2 instance role through IMDSv2, and neither promise was true.

Not a consumer: the S3 audit sink. `ee/engine/auditsink/configure.go:63` reads `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` itself and calls `cloudauth.SignV4` directly, so it never reaches instance metadata. It cannot use an instance role at all. That is a separate limitation, and this change does not touch it.

## How it was found

On 2026-09-13 the Aurora provider ran against real AWS for the first time, on a t4g.micro in the source cluster's VPC. The instance required IMDSv2 with hop limit 1 and carried an instance role with the IAM actions the Aurora page documents. `aurora.New` failed in 1.2 seconds with "no AWS credentials were found ... instance metadata did not answer".

In the same empty environment on that instance, curl fetched the token, the role list and the credentials document, each in under three milliseconds. A probe calling `cloudauth.Do` for the token got 200 in 1.4 ms. The full chain still said nothing had answered.

lane-rds hit the same defect on its own runner, i-0216ddedf2e6d0078, a t4g.nano with HttpTokens required, hop limit 1, and instance role af-rds-proof-runner. It read only status codes and printed no credential values:

- The token PUT, sent as Go sends it, got 200 in about 1 ms.
- The role listing with the token returned the role.
- The credentials GET with only `Accept: application/json`, which is exactly what `fromEndpoint` sent, got 401.
- The same GET with the token got 200.
- At 15:16:20Z, `rds.Registration{}.Open` failed after 3 ms with the same "instance metadata did not answer".

lane-rds ruled out the air gap guard (not sealed), proxy variables (none set) and the network (the same PUT from the same host answered).

## Confirmed live with this fix, on a second instance role

lane-rds ran its RDS provider against this change on its own runner, i-0216ddedf2e6d0078: IMDSv2 required, instance role af-rds-proof-runner.

- **The binary.** Built natively for linux/arm64 on a separate build instance, from #400's head 4746b8b9 plus lane-rds's rotation fix plus this PR's `aws.go` and `aws_test.go` at 6d9490bc, as lane-rds states it. Its sha256 was 5d0d96d71dc7f5fe5d70031dd255712741b3cf142e58f184f4b2c048998d72fd on the builder, on the Mac and on the runner.
- **The environment.** The run started at 16:11:53Z. The script refuses to start unless no AWS key variable is present, and the runner's log records "AWS key variables in the environment: 0". So the credential chain had to reach instance metadata.
- **The result.** "open the provider through its registration (describes the source) took 0.2s, error <nil>". That open is a SigV4 signed DescribeDBInstances made with the instance role's credentials.
- **Before the fix.** The same step on that runner failed at 15:16:20Z in 0.003 seconds with "instance metadata did not answer".

The evidence is lane-rds's `orchestrate-3.log` and `run.log`, which record the start, the empty environment and the result line with timestamps.

## Why nothing caught it

The metadata address was a constant, so no test ever ran `fromInstanceMetadata`. `TestAWSChainNamesEveryPlaceItLookedForCredentials` checks only the final refusal. On an EC2 instance with a role, the real address answers, and that test passed there only because the chain could not read the role it had found.

## The change

- `AWSChain` carries the metadata base address. `NewAWSChain` sets it to 169.254.169.254, and only a test changes it.
- `fromEndpoint` takes a header map. The ECS paths pass Authorization when a token is set, and instance metadata passes the IMDSv2 token.
- `fromInstanceMetadata` reports whether metadata answered. If it did and the role path then failed, the refusal says so and gives the reason, for example `answered 403`. A machine where the token request fails still gets the original message naming every place the chain looked.

## Tests

The fake behaves as IMDSv2 does: a GET without the session token is answered 401, the credentials document included.

- `TestAWSChainReadsAnInstanceRoleThroughIMDSv2` checks the access key, session token, expiry, and a source naming the role.
- `TestAWSChainSaysWhatAnInstanceRoleAnsweredRatherThanThatNothingDid` requires the 403 to be named and "did not answer" to be absent.
- `TestAWSChainNamesEveryPlaceItLookedForCredentials` now points at a server that has already closed.

## Mutation

These cells ran on the proof runner (linux/arm64, go1.26.8) in a scratch copy of main plus this change. Each cell ran its named test exactly once and then restored it, and afterwards the tree matched its pristine sha256.

- Passing nil instead of the token on the credentials read fails `TestAWSChainReadsAnInstanceRoleThroughIMDSv2`.
- `fromEndpoint` ignoring the headers it is given fails the same test.
- Skipping the answered branch fails `TestAWSChainSaysWhatAnInstanceRoleAnsweredRatherThanThatNothingDid`, and its message then quotes "did not answer".
- Removing the closed-server address fails `TestAWSChainNamesEveryPlaceItLookedForCredentials` on the EC2 instance, where the real metadata service now hands out the role.

## What ran where

- **On the proof runner** (i-0eee396e1f9cec69c: linux/arm64, go1.26.8 pinned by `go.mod`, tarball sha256 checked against go.dev), using a scratch copy of main plus this change. The whole `cloudauth` suite passed, `go vet ./cloudauth` passed, and the four mutation cells above ran. The files in this commit are byte-identical to the copies tested there.
- **Not run on this machine before opening.** The shared build lock queue was seven deep. lane-rds needed a head to confirm against on its own runner before tearing that runner down, so this opened without the local gate. The gate is queued: gofmt, go vet on `cloudauth` and `secrets`, and go test on `cloudauth`, `secrets` and `db/aurora`. Its result will be posted here. CI's engine and enterprise jobs cover the same packages.
- **Live.** With this fix in the test binary, the Aurora provider got past `aurora.New` on the instance role and cloned a golden from a real Aurora cluster. That run then stopped on three separate Aurora defects, which go in their own pull request, because #374 is rewriting the same provider files.


VirSanghavi:
Local macOS gate, run after merge as promised in the body.

It ran on 6d9490bc9defb719f34811dc9987549d79ed8759, the head that merged as 2d3174f3, under the shared build lock (acquired 16:24:47Z, finished by 16:27:41Z), with `GOWORK=off`:

- `gofmt -l cloudauth`: nothing listed
- `go vet ./cloudauth ./secrets`: exit 0
- `go test ./cloudauth ./secrets ./db/aurora -count=1`: exit 0
  - `ee/engine/cloudauth` ok in 2.249s
  - `ee/engine/secrets` ok in 40.757s
  - `ee/engine/db/aurora` ok in 160.188s

This agrees with the proof runner's linux/arm64 run and with CI, where all nine required contexts, Antifailure and dogfood read success on this head.

PR #421, opened 2026-09-13
throwaway: dogfood never runs when a draft is marked ready

VirSanghavi:
Throwaway. Not for merge, and it will be closed with its branch deleted.

It carries main's own `.github/workflows/dogfood.yml` on an empty commit, to reproduce the defect before the fix: the control plane starts no check on a draft and opens one on ready for review, while dogfood.yml only runs on opened, synchronize and reopened. So this draft's run cannot report, and marking the pull request ready is expected to start no dogfood run at all.

The fix is then pushed here, to show a draft skipping the check with a comment saying so, and marking ready starting a run that goes green.


VirSanghavi:
Closing. This was the reproduction for #422, not a change: an empty commit carrying main's own dogfood.yml, then the fix pushed onto it, so the draft to ready transition could be watched on both sides. All four transitions and their run ids are quoted in #422, which has merged. The branch is deleted.

PR #512, opened 2026-09-20
engine: the desktop driver was proven, shipped, and no manifest could ask for it

VirSanghavi:
Stacked on #509, which is the driver. This is the half that makes it reachable.
Merge #509 first; this branch is #509 plus one commit.

## The failure

The desktop driver reached the runner's job document and stopped there. Nothing
in a manifest could name a desktop application or a desktop workflow, so the
engine sent the runner a document with none in it on every run that has ever
happened.

A driver nobody can reach is not a feature, and it is the defect the terminal
list was added to close one surface over. Finding the same shape again
immediately is the part worth noticing: the driver and the manifest are written
by different hands, only the manifest is what a customer can see, and nothing
between them fails when the second one is missing.

## What this adds

`desktop` says which application the workflows drive: its kind, `electron` or
`macos`, and the path to it. `desktop_workflows` is what the agents do in it,
written exactly as a browser workflow is: a sentence saying what a person would
do and what proves it happened. They run in the same `af test` against the same
environment, obey the same `--only`, share one namespace with the other two
lists because a name is what a report prints, and are counted, judged and
reported exactly like a browser result.

It follows the terminal list's pattern rather than inventing a second one: a top
level list, a builder in the orchestrator, absolute path resolution before the
document is sent, and the surface decided in `surfaceFor`.

## Why a second list rather than a `surface` key on `workflows`

The terminal list's reasoning, holding here for a different pair of fields. A
browser workflow needs a persona to sign in as and a path to start at, and a
desktop application has neither: there is no address bar to open and no cookie
to set, so signing in is itself a workflow that types into the fields the
application shows. One entry carrying both halves would refuse half its own keys
on every manifest.

## The count that would have made all of it inert

`Test` refuses a run that declares no workflows, and that guard added two of the
three lists. A manifest whose only workflows are desktop ones was therefore
refused as declaring none: the schema would have accepted it, the validator
would have passed it, the documents would have been built, and the run would
have stopped one line before any of it mattered.

That is the same defect as the driver's, one layer up, which is why this change
asserts the count and not only the transport.

## What the surface string means

Said once, because two lanes have now had to decide it. It answers exactly one
question: is a browser opened.

Terminal workflows run whatever it says, because a command line program needs
nothing opened for it, and desktop workflows do too, because the runner
dispatches all three lists and only the browser is gated on this string. So a
run with browser work stays a web run and drives its desktop workflows alongside
it, and a run with no browser work says `desktop` so that no browser is opened
for a run with no page to open. Neither surface can starve the other.

## A rule this deliberately does not copy

The terminal list refuses an expectation the workflow also types, because a
pseudo terminal echoes its input and such an expectation is satisfied by the
workflow rather than by the program. That question was asked of this surface
too, and the answer is different.

The same hole existed and was closed one layer down, in
`runner/src/drivers/ax.ts` on #509: a field's own value is left out of the text
expectations are judged against. With the value excluded, an expectation naming
an answer can only be met when the application RENDERED those words, and a
confirmation screen reading back the address somebody typed is exactly the
evidence such a workflow is written to find. Copying the rule here would refuse
a correct workflow, so the absence is asserted by a test rather than left to be
rediscovered and "fixed".

## The gate that reports red while measuring nothing

`TestEverySchemaConstraintIsEnforced` needed its tuning in the same commit, for
the second feature running. A list whose entries share a namespace with another
list has its generated name collide with that list's, and a block with a
required enum is generated without one. Both refuse the base manifest, and a
refused base means every one of the 787 cells measured NOTHING while the gate
reported red, which reads exactly like a bug in the feature. It caught this
honestly and said so in those words. The note above `wantConstraints` now states
it as a rule rather than as a story, because it has now happened to two features
in a row.

## Proof

* **Mutation table: 12 cells, 12 red, 0 survivors**, across the validator, the
  normaliser and the orchestrator: both halves of the application and workflow
  pairing, the `process` refusal, names shared with the terminal list, the
  description floor, the step default, the derived process name, absolute path
  resolution, the application withheld when nothing drives it, the `--only`
  filter, the step budget reaching the runner, and the desktop surface being
  named at all.
* One cell survived first and was re aimed rather than accepted. Its anchor,
  `doc.MaxSteps = w.Budget.Steps`, appears TWICE in `test.go`: once in
  `workflowDocs` for the browser list and once in `desktopDocs`. The harness
  replaced the first, so it mutated the browser's line, the desktop test passed
  correctly, and the cell reported SURVIVED while having measured nothing about
  the code it named.
* `internal/manifest` green, including the 787 constraint gate. `pkg/schema`
  green. The desktop tests in `internal/env` green.
* `fieldsweep` accounts for 299 of 299 manifest fields with the new ones read.
* `prosecheck` 1148 files 0, `defercheck` 3434 files 0 deferrals, `changecheck`
  ok.

## What could not be checked here

The machine's screen is locked, which blocks two things that have nothing to do
with this change and everything to do with the host.

`go test ./internal/env` as a whole times out: it hangs in `SystemKeyring.Get`,
which shells out to `security`, which a locked Mac will not serve. The desktop
tests in that package pass on their own.

`gendrift -strict` cannot produce a receipt, because `go test ./internal/cli
-update-reference` times out at ten minutes on this host. The generators this
change actually feeds all ran and are settled: `schemadoc`, the copy of the
schema into `engine/internal/manifest/`, `proxysrc` and `docsembed` were re run
and a second pass changed nothing, and the two schema copies are byte identical.
`cli.md` is untouched because this changes no command.


VirSanghavi:
Superseded by #525, which is the same capability minus the part of the design this one predates.

#512 proved the desktop driver and shipped it, and the manifest could not ask for it. The ruling that superseded this branch is that the request belongs in `workflows[].surface` beside every other surface rather than in a `desktop_workflows` list of its own, because a second list is a second place for the count of workflows to be right, and `Test`'s own guard reads one of them. #525 carries no `desktop_workflows` and no deprecated alias, and it needed no change to the runner at all: `main.ts` already routed `surface === 'desktop'` to `runDesktop`.

Closing rather than leaving it open, so nothing of the old shape merges by accident. The branch is not deleted.

